PR #2, opened 2015-06-14
Fixed django 1.8 allow_migrate router signature, documentation

ivome:
The signature of allow_migrate has changed in django 1.8. See:
https://docs.djangoproject.com/en/1.8/topics/db/multi-db/#allow_migrate
With the current implementation, the migrations do not work as expected. I ran into issues when I had a shared app with a migration that had a RunPython operation. That operation was executed on the tenant on creation as well as on migrate_schemas and failed obviously because of missing database tables. 
With the small update it runs fine again. 
I also removed some references to the removed sync_schemas commands in documentation.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -160,12 +160,12 @@ created ``Client`` inside an app named ``customers``, your
 
     TENANT_MODEL = "customers.Client" # app.Model
 
-Now run ``sync_schemas``, this will sync your apps to the ``public``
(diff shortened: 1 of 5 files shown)


tomturner:
Thanks for this.

tomturner:
It on pypi now so you can pip install it

ivome:
Wow, that was quick! Thanks a lot! :+1: Great project btw!

PR #10, opened 2015-06-28
Prevent potential infinite recursion in DomainMixin.save().

suriya:
Rewrite DomainMixin.save() in such a way to prevent a potential infinite
recursion. Previously, the save() method invoked save on other DomainMixin
objects. An infinite mutual recursion could occur with two objects, both
with is_primary=True, calling each other's save() methods.

Not sure of the exact conditions that could lead to this behavior.
Nevertheless, better to prevent it.

### DIFF
--- a/django_tenants/models.py
+++ b/django_tenants/models.py
@@ -1,7 +1,6 @@
 import django
 from django.conf import settings
 from django.contrib.auth import get_user_model
-from django.contrib.postgres.fields import ArrayField
 from django.db import models, connection, transaction
 from django.core.management import call_command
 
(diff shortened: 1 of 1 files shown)


django_tenants/models.py
```diff
@@ -127,16 +126,11 @@ class DomainMixin(models.Model):
...
-                # Remove primary status of existing domains for tenant
-                for domain in domain_list:
-                    domain.is_primary = False
-                    domain.save()
-
+        # If we have no primary domain yet, set as primary domain by default
+        self.is_primary = self.is_primary or (not domain_list.exists())
+        if self.is_primary:
+            # Remove primary status of existing domains for tenant
+            domain_list.update(is_primary=False)
```
gannetson: If you're trying the to save a domain that's already has `is_primary=True` this command will set it to `False` and then the super in the next line will set it to `True` again, right? Or should we do something like 
`domain_list.exclude(id=self.id).update(is_primary=False)`

suriya:
`domain_list` already has the `exclude()` that is required.

```python
domain_list = self.__class__.objects.filter(tenant=self.tenant, is_primary=True).exclude(pk=self.pk)
```

I wanted to rename it to `other_domains`, but did not want to change the name that the original author used. You can change the variable name to `other_domains`, and it will make the code easier to read.

suriya:
`super()` is called only for the domain that is being saved.

gannetson:
Cool thanks!

PR #11, opened 2015-06-28
Fix domain creation bug in create_tenant management command.

suriya:
The 'create_tenant' management command was not handling domain creation
options properly and was running into an infinite loop. This was because
the store_tenant_domain() method was not returning the domain that it had
newly created.


gannetson:
Thank you!

PR #15, opened 2015-07-06
Syntax sugar for tenants

kosz85:
Hi,
I'm using overpatched version of django-tenant-schemas for commercial project for about 2 years, I'm now porting it to django 1.8 so I'm starting using django-tenants.
I'm using it quite heavily, so I want to share with you some syntax sugar that I forged for my project. I love your clean approach to tenants, it seems much better then previous project. Though I think this little additions can be really helpful, especially for some folks not knowing python so well.

Details in the docstrings.

I can also try share celery tasks that remembers schemas that are run from, and features like that. If you want I can be adding them while I will be porting app. In future I will also need some parallel schema migrations that can be shared, as when number of tenants grows it starts to be really slow to migrate them sequentially.

### DIFF
--- a/django_tenants/models.py
+++ b/django_tenants/models.py
@@ -34,6 +34,42 @@ class TenantMixin(models.Model):
     class Meta:
         abstract = True
 
+    def __enter__(self):
+        """
+        Syntax sugar which helps in celery tasks, cron jobs, and other scripts
+
+        Usage:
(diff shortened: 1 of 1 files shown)


tomturner:
Please share as much as possible. I agree about updating the migrations in parallel. I need a patch for that!

PR #21, opened 2015-07-17
fix for django 1.8

tramwaj29:
app_labels() was needed for old django versions, it was causing problems for django 1.8
e.g. 'kombu.transport.django' was converted to 'django' and obviously not handled. Since we don't want app_lables() anymore we need app_name


tomturner:
Thanks for this update

PR #22, opened 2015-07-22
Adding support for PostGIS with minimal code changes

george-silva:
Hello @tomturner. I'm making this patch because I do need to use this with PostGIS. Since this is already tied to PostgreSQL, I don't think we need to cover other backends.

I've added a minimal documentation section, but I really don't the format (I know more markdown). I just tested this on a brand new project, following the instructions we have on the documentation and on the ones I've added and it's working.


tomturner:
Thanks you very much I will try it later

PR #39, opened 2015-10-07
Added migration executor layer

### LINKED ISSUE #38: Parallel migrations
We need a way of doing Parallel migrations

CGenie:
- standard executor preserves current behavior (i.e. applies tenant migrations
  in sequence)
- multiprocessing executor applies tenant migrations asynchronously

Fixes #38

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -219,9 +219,15 @@ You can run tenant migrations in parallel like this:
     python manage.py migrate_schemas --executor=multiprocessing
 
 In fact, you can write your own executor which will run tenant migrations in
-any way you want, just take a look at `django_tenants/migration_executors`.
+any way you want, just take a look at ``django_tenants/migration_executors``.
 
+The ``multiprocessing`` executor accepts the following settings:
 
+* ``TENANT_MULTIPROCESSING_MAX_PROCESSES`` (default: 2) - maximum number of
+  processes for migration pool (this is to avoid exhausting the database
+  connection pool)
+* ``TENANT_MULTIPROCESSING_CHUNKS`` (default: 2) - number of migrations to be
+  sent at once to every worker
 
 Credits
 -------
--- a/README.rst
+++ b/README.rst
@@ -209,6 +209,19 @@ tutorial, such as creating the public tenant and configuring shared and
 tenant specific apps. Complete instructions can be found at
 `django-tenants.readthedocs.org`_.
 
+Parallel migrations
+-------------------
+
+You can run tenant migrations in parallel like this:
+
+::
+
+    python manage.py migrate_schemas --executor=multiprocessing
+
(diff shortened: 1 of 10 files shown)


ivome:
Really nice feature!!

One question: Couldn't you easily run into issues with too many database connections, when your number of tenants exceeds your max connections on the DB server? Maybe it would be safer to be able to limit the level of concurrency, also to not get too many fork processes at once.

CGenie:
Sure, I'll add such limit as a django settings.

tomturner:
Really Good

in django_tenants/migration_executors/base.py you are checking that the version is greater than 1.7. You can get rid of that line

Any chance of a test for this feature.

CGenie:
Yeah, I started to develop for https://github.com/bernardopires/django-tenant-schemas then I noticed that we use your fork so I forgot that 1.7 :) Removing.

I'll add tests.

tomturner:
thanks

CGenie:
I added option to specify custom executor when running tests, they seem to pass both with default `standard` executor and with the `multiprocessing` one. I can also add unit tests for the executor module if you want, it's just I don't see unit tests here at all :)

tomturner:
@CGenie Testing is a bit thin as the parent project didn't have a lot. I am now trying to get untests a bit better.

CGenie:
I actually looked more carefully at tests -- I don't see point of inserting unittests to this code as it is quite simple. As for testing if migrations work, we have such tests in `test_tenants.py` so it is sufficient to run tests for every executor I think. Do you agree?

CGenie:
If you want I can add Travis integration to the project :)

tomturner:
@CGenie I will get this merged in later today thanks for your hard work

CGenie:
Tests pass, cool that this will be merged :)

kosz85:
Nice :) Thx, for merging. I thought it wold take more time. @CGenie is a real code genie ;)

tomturner:
@kosz85 I guess he works with you. I can see he is very good coder

alexandrebarbaruiva:
@CGenie is there any way to add custom executors on my project and use it on migration?

CGenie:
@alexandrebarbaruiva Looking at
https://github.com/tomturner/django-tenants/blob/master/django_tenants/migration_executors/__init__.py
it seems you can. Just add a new class with a custom codename like this one:
https://github.com/tomturner/django-tenants/blob/master/django_tenants/migration_executors/standard.py
I don't know what you want to do in your custom executor, but task here was to wrap whatever `django-tenants` did during migrations into parallel code. So it's just about calling `run_migrations` from
https://github.com/tomturner/django-tenants/blob/master/django_tenants/migration_executors/base.py
which is a wrapper around `django`'s `MigrateCommand`.

PR #49, opened 2015-11-10
Fix example in Documentation

suriya:
While setting up our test case, invoke `TenantTestCase.setUp()`. Without this, we don't get any of the benefits of `TenantTestCase`.


tomturner:
Fixed #49 thanks @suriya

PR #51, opened 2015-11-11
Fix check of an app's presence in INSTALLED_APPS

suriya:
In TenantSyncRouter, the logic to check whether an app is a tenant app or shared app was too simplistic. Django 1.7 allows two ways to add an app to INSTALLED_APPS. 1) By specifying the app's name, and 2) By specifying the dotted path to the app's AppConfig's class. This commit ensures that we check for the latter case as well. https://docs.djangoproject.com/en/1.8/ref/applications/#configuring-applications

### DIFF
--- a/django_tenants/routers.py
+++ b/django_tenants/routers.py
@@ -1,4 +1,5 @@
 from django.conf import settings
+from django.apps import apps as django_apps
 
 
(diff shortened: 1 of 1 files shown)


tomturner:
@suriya Thank you

PR #65, opened 2016-01-12
TemplateDoesNotExist has been moved in django

Pesticles:
The TemplateDoesNotExist exception has been moved from django.template.base to django.template.exceptions; changed the import.


tomturner:
What version did it django change this on?

As it need to work on 1.8 and 1.9

ka7eh:
From 1.9; here's the related commit:
https://github.com/django/django/commit/d17a035132c4ddba51d57286b2185bc459f69067

PR #74, opened 2016-03-18
fix url for help

antoniocdff:
Changed url to point a wrong location


tomturner:
Thanks must of missed that one

PR #75, opened 2016-04-25
Develop

incognos:


tomturner:
Did you mean to cancel this pull request

incognos:
Yes I did… not ready for primetime :-)

Trying to do a multi-db implementation where the public schema routes to one db and the tenants route to a different db.

I also, am trying to add the ability to select schema based on a session variable

Cheers,
Sina

Issue #76, opened 2016-04-28
Typo in docs - django-tenants-schema 


bartmika:
Hi,

When I look here: http://django-tenants.readthedocs.io/en/latest/use.html

I see you have **django-tenants-schema**, should it not be **django-tenants**?

tomturner:
Yes I will change it unless you want to put a pull request

bartmika:
I'd like to make a pull request.

tomturner:
No problem you can do so

PR #79, opened 2016-04-28
Typofix #76: Removed the old text from the **django-tenants-schema** …

bartmika:
Here is the typofix for issue: https://github.com/tomturner/django-tenants/issues/76


tomturner:
Thanks

Issue #80, opened 2016-04-28
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 604


bartmika:
Hello, I am running **Python 3.4** on a FreeBSD box and I ran the following command:

```bash
pip install django-tenants
```

But unfortunately it errors, here's the output:

```
Collecting django-tenants
  Downloading django-tenants-1.1.5.zip (87kB)
    100% |################################| 92kB 2.5MB/s
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-kblbhq12/django-tenants/setup.py", line 31, in <module>
        long_description=open('README.rst').read() if exists("README.rst") else "",
      File "/usr/home/django/py-bizmula/env/lib/python3.4/encodings/ascii.py", line 26, in decode
        return codecs.ascii_decode(input, self.errors)[0]
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 604: ordinal not in range(128)

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-kblbhq12/django-tenants/
```

Any way we can fix this? Thanks!

rodolfomartinez:
+1

viralrain:
+1

kosz85:
For python 3.5 it's working, so I checked Python 3.4.4 and...  ...it's also working?
```
Python 3.4.4 (default, Mar  7 2016, 03:46:01) 
[GCC 5.3.1 20160225] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import django_tenants
>>> dir(django_tenants)
['__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', 'default_app_config']
```
I think it's smth internal in your python interpreter minor version or system. Please check all versions, I installed it with:
```Python 3.4.4
pip 8.1.1```
Provide your versions of python, pip, and system itself.

bartmika:
I have two machines running two different OSes and here are the info:

**OS X**
Version - 10.11.4
Python 3.4.2
pip 1.5.6 
*django-tentants works*

**FreeBSD**
Version - 10.2
Python 3.4.4
pip 8.1.1 
*django-tentants does not work*

bartmika:
[...]

PR #81, opened 2016-04-30
BUGFIX: Fixed setup.py file to force utf-8 encoding for setup in case…

bartmika:
… an OS which is trying to install django-tenants and is using ascii encoding thus preventing any ascii errors. This fixes issue #80 .

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ setup(
     url='https://github.com/tomturner/django-tenants',
     license='MIT',
     description='Tenant support for Django using PostgreSQL schemas.',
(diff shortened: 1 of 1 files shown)


tomturner:
This merge would break python 2.7 so I can't except at the moment


Traceback (most recent call last):
  File "setup.py", line 31, in <module>
    long_description=open('README.rst', encoding='utf-8').read() if exists("README.rst") else "",
TypeError: 'encoding' is an invalid keyword argument for this function

bartmika:
My bad.

Issue #87, opened 2016-07-04
"reverse" with tenant?


bartmika:
How do I utilize the django urlresolve **reverse** function to take into account a particular tenant subdomain in the URL? I didn't see any custom **reverse** function in https://github.com/tomturner/django-tenants/blob/master/django_tenants/urlresolvers.py

Ex:

``` python
from django.core.urlresolvers import resolve, reverse
url = reverse('tenant_dashboard')
print(url)  # Result: /en/dashboard

# The next code is the desired functionionality I am looking for
url = tenant_reverse("some_tenant_name", "tenant_dashboard")
print(url)  # Result: http://some_tenant_name.mydomain.com/en/dashboard
```

I ask this question because I noticed the `django-subdomains` library has this sort of functionality. See here: https://github.com/tkaemming/django-subdomains/blob/master/subdomains/utils.py#L41

 Thanks!

rodolfomartinez:
+1

bartmika:
Alright, my fix was this code:

```python
from django.contrib.auth.models import User
from django.contrib.sites.shortcuts import get_current_site
from django.core.urlresolvers import resolve, reverse
from foundation_public.models.organization import PublicOrganization

# ...

def index_page(request):
    org = PublicOrganization.objects.get(owner_id=request.user.id)
    url = 'https://' if request.is_secure() else 'http://'
    url += str(org.schema_name) + '.'
    url += get_current_site(request).domain
    url += reverse('tenant_dashboard')
    print(url)  # Result: http://my_test_tenant.my_test_domain.com

# ...
```

However, I think having a custom reverse function for specific tenants would be immensely useful.

tomturner:
I can add that in at some point. If someone create a pull request tat would be good.

kosz85:
What a problem adding such func in your project? Everyone can do it on it's own, as not everyone is using sites ;) For example I have custom reverse so it's adding me `URL_PREFIX` which is smth like `/tenant/{{tenant.slug_name}}` so finally it's joined into smth like `/tenant/abc/some/url/in/app`, and I have many tenants per one domain. It all depends how you use this lib.

bartmika:
Hmmm you make a good point @kosz85. What do you think we should do in this case? I think having somesort of **resolve** function per tenant to be a extremly useful function.

kosz85:
I was thinking a bit about this issue. For me it's more like thing for examples chapter in docs (some kind of HowTo), then smth placed in core lib.
[...]

PR #89, opened 2016-07-05
Added reverse for tenants. #87

bartmika:
The following code grants the Tenant models (which inherit **TenantMixin**) to reverse views to specific URLS with sub-domains handling using Django's existing Sites Framework. Code for issue #87.

### DIFF
--- a/django_tenants/models.py
+++ b/django_tenants/models.py
@@ -1,6 +1,8 @@
 from django.conf import settings
 from django.db import models, connection, transaction
 from django.core.management import call_command
+from django.contrib.sites.shortcuts import get_current_site
+from django.core.urlresolvers import resolve, reverse
 
 from .postgresql_backend.base import _check_schema_name
 from .signals import post_schema_sync, schema_needs_to_be_sync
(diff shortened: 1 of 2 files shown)


django_tenants/models.py
```diff
@@ -145,6 +147,16 @@ def get_primary_domain(self):
...
 
+    def reverse(self, request, view_name):
+        """
+        Returns the URL of this tenant.
+        """
+        url = 'https://' if request.is_secure() else 'http://'
+        url += str(self.schema_name) + '.'
+        url += get_current_site(request).domain
+        url += reverse(view_name)
+        return url
```
kosz85: `''.join()` is faster then `+` which is constructing intermediate strings, that takes memory and slows down process of building url. So it would be better to have here `''.join((protocol, self.schema_name, '.', domain, reverse(view_name))`

django_tenants/models.py
```diff
@@ -145,6 +147,16 @@ def get_primary_domain(self):
         except get_tenant_domain_model().DoesNotExist:
             return None
 
+    def reverse(self, request, view_name):
+        """
+        Returns the URL of this tenant.
+        """
+        url = 'https://' if request.is_secure() else 'http://'
```
kosz85: It's more then reverse is doing, I would suggest to leave just `//` and leave deciding the protocol by browser,  tought mail programs can have problem with such urls, as they don't have current used protocol. Also what if reverse is used in an email, so some notification sent by celery, without request? This shouldn't be looking at request, or request should be optional.

django_tenants/models.py
```diff
@@ -145,6 +147,16 @@ def get_primary_domain(self):
         except get_tenant_domain_model().DoesNotExist:
             return None
 
+    def reverse(self, request, view_name):
```
kosz85: I would rather leave view_name on first position, to not be changing API of this method, and made request optional keyarg

tomturner:
@bartmika If you do the changes that @kosz85 suggests I will accept your changes

PR #108, opened 2016-11-19
Specializing static and media based on tenants

mateuspadua:
New Feature to be possible have **different static files for each tenant or upload media files to separate directories for each tenant**. More information, please read the documentation on this commit. :)


kosz85:
Nice addition, but... From my personal story :) Using `schema_name` in `path` can be a bit problematic, especially when you can (as app feature) or have to (as client request) rename `schema_name`. I have less advance solution of this (just constructing path where it's needed), but really using `tenant.id` is much better option, as even after renaming it stays the same. It could be an option if you want to use `schema_name` or `id`, it's small change but it will save you problems in future.

mateuspadua:
Thank you for your considerations. I followed the same idea of what already existed for the templates based on tenants, so I used schema_name, but later, we can put a flag in the settings where we can choose between using ID or schema_name. :). Thanks

django_tenants/middleware.py
```diff
@@ -26,7 +32,9 @@ def process_request(self, request):
         domain = get_object_or_404(get_tenant_domain_model().objects.select_related('tenant'),
                                    domain=hostname)
 
+        domain.tenant.domain_url = hostname
```
tomturner: @mateuspadua Is this line needed as i don't think it is

mateuspadua: I dont add the line 35, this line already existed, see here. :)
https://github.com/tomturner/django-tenants/blob/master/django_tenants/middleware.py

mateuspadua:
Hi @tomturner, i'm here if you need any help to merge my pull request. Thanks

tomturner:
@mateuspadua Thank you I will try and merge over the weekend

mateuspadua:
@tomturner thanks for merge, i believe that it is going to help many people. :)

Issue #114, opened 2017-02-12
Custom command ignoring tenant


perrohunter:
Hi, I'm writing a custom command and I'm realizing that even if I specify a `--schema` when running the custom command with `tenant_command` some times I'm getting error:

```
django.db.utils.ProgrammingError: relation "my_table" does not exist
LINE 1: SELECT (1) AS "a" FROM "my_table" WHERE "my_tabl...
```

I've notice that this happens when I'm trying to access a `tenant` app and suddenly some models are not properly tenant configured and throw the error.

Any away I can ensure the right schema is properly configured?

This **only** happens when running code through a `tenant_command` where I'm loading modules dynamically and one of them causes an exception.

My Logic:

1. run `./manage.py tenant_command demodata`
   my `demodata` command looks through all the django applications, find the ones containing a `demo` folder and loads each file in it the calls a `install()` method [sample gist](https://gist.github.com/perrohunter/ee6ef1fe49c9876a9d90ad33f76ba388)

2. `demodata` command files two files `cats.py` and `dogs.py`, it loads them and then calls a method first on `cats.py` then `dogs.py`
3. cats.py fails with an exception "integrity bla bla, duplicate key" so far, everything is as expected
4. dogs.py gets run and upon executing the first db query `Dogs.objects.filter(goodBoy=True)` it fails with 

```
django.db.utils.ProgrammingError: relation "pets_dogs" does not exist
LINE 1: SELECT (1) AS "a" FROM "pets_dogs" WHERE "pets_dog...
                               ^
```

this is a really weird scenario, if no exception is caused by `cats.py` then `dogs.py` runs just fine

perrohunter:
So I found the cause, for some reason when I cause a exception on the database side, let's say an Integrity Error which raises a `django.db.utils.IntegrityError` for some reason a new cursor gets created on the django side, this new cursor doesn't have the proper `search_path` but the `_cursor()` funcion on `base.py` things it does and therefore it doesn't re-set the `search_path` a way around this is to monitor if the cursor we are getting is different from the previous cursor we got.

New attribute of the `DatabaseWrapper`
```
_previous_cursor = None
```
and inside _cursor() validation of `search_path`
```
if (not get_limit_set_calls()) or not self.search_path_set or self._previous_cursor != cursor:
            self._previous_cursor = cursor
```

tomturner:
@perrohunter Thank you I will have a look at the pull request

tomturner:
Merged

PR #115, opened 2017-02-13
Fix for #114

perrohunter:
This fix introduces a validation of the cursor at the `DatabaseWrapper` to make sure the search_paths are always applied even if we get a different cursor pointer


tomturner:
@perrohunter Thanks

PR #117, opened 2017-02-21
FastTenantTestCases also call TenantTestCase

CodeMonk:
This fixes a bug in the FastTenantTestCase where the instance
setup/teardown functions were being called, as well as the class level
setup and teardown. This was causing inconsistency failures as the same
records were attempting to be written twice.


perrohunter:
kudos

PR #137, opened 2017-08-09
Introspection fix

CGenie:
I copied `get_constraints` from `django.db.backends.postgres.introspection` (https://github.com/django/django/blob/1.11.4/django/db/backends/postgresql/introspection.py#L151-L248).
This was because `migrate_schemas` threw error about missing `'type'` key in a constraint dict.
This shouldn't break earlier versions of Django as they just added a couple of keys to the dict.

I moved `get_constraints` function into `_constraints.py` file and versioned it according to Django 1.11 or earlier. Basically we are migrating to 1.11 so this version is more important to us than older ones :)

### DIFF
--- a/django_tenants/postgresql_backend/_constraints.py
+++ b/django_tenants/postgresql_backend/_constraints.py
@@ -1,0 +1,199 @@
+import django
+
+if django.VERSION < (1, 11):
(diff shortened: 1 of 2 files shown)


tomturner:
Thanks for that. I will do a PyPi release tomorrow.

PR #138, opened 2017-08-14
Fix django 1.9 and 1.10 problem with the commands

yamijuan:
In the `clone_tenant.py` command there was a problem with `BaseCommand.option_list` that doesn't exists in django 1.10.+ . Also, when the commands were loaded by some method like `django.core.management.load_command_class` using `get_commands` from the same module, if the version was not django 1.8.x or less, the `syncdb.py` command died because the django syncdb command was removed in django 1.9.


tomturner:
Thank you very much. I will check it works on Django 1.11 then I will merge it

PR #139, opened 2017-08-17
Percent format

CGenie:
Just added a display of count numbers when outputting tenant schema with percent of tenants migrated. This makes waiting for migrating a long tenant list easier. :)


tomturner:
Thanks again

PR #143, opened 2017-10-18
Fix django.urls.reverse import DeprecationWarning

suriya:


webjunkie:
@tomturner Can this be merged? It seems django-tenants is not Django 2.0 at this time.

suriya:
I thought djangol-tenants was Django 2.0 ready with commit fd1996e115072485c158988e591159c89f4ef42b.

@tomturner I would prefer that django-tenants support all Django LTS releases that are currently active. At this moment, 1.8 and 1.11 are the active LTS releases. Django 1.8 is supported until April 2018. I notice that you have removed 1.8.

Also, looks like django-tenants does not support Python 2.7 (from the setup.py). Ideally, we should support Python 2.7 and Django 1.11 until April 2020.

tomturner:
The next version of Django tenant will only Django 2. This is what Django recommend in there release notes. If anyone wants to use Django tenants on the other versions they can stay with the version that is already released. The reason is that this means lots of check of version can go which will speed the code up and make it easier to maintain. I will be working on get this new version within a week of now.

suriya:
From
https://docs.djangoproject.com/en/2.0/releases/2.0/#third-party-library-support-for-older-version-of-django

> Following the release of Django 2.0, we suggest that third-party app
authors drop support for all versions of Django prior to 1.11.

I suggest that django-tenants supports Django 1.11 LTS as well. I
understand the relief you will get from cleaning up code. I think it would
be prudent to continue supporting the current LTS. Django 1.11 has a longer
life than Django 2.0.

kosz85:
Then maybe create branch that support django 1.x and django 2.x, then add info to Readme about which version use with what Django. And up django-tenants version to 2.0 for django 2.x.
There may be still people using it with python 2.7, and there still may be needed some fixes.

suriya:
My suggestion is to keep one branch and support both versions, which is the current practice. Continuing this until Django 1.11's end-of-life should be straightforward.

tomturner:
I understand people wants Django Tenant to support both Django 1.X. However I don't want the code not to be a mess. That was the problem with the code before I forked the project. As I said before this project will drop support for Django 1.X. You can still use version 1.3.3 which supports Django 1.X.
If a major bug is found in Django tenant in version 1.3.3 then I would consider creating a branch at that point.

tomturner:
This has been fixed in my Django 2 support

suriya:
@tomturner I am currently running Django 1.11 LTS with django-tenants 1.3.3. My practice for upgrading Django is to go from one LTS release to the next. Before upgrading, I ensure that I handle all `DeprecationWarning`s. (In fact, that's what prompted this pull request).

django-tenants 1.3.3 supports Django <= 1.11 but not Django >= 2.0. The next release django-tenants 2.0.0 supports only Django >= 2.0 but not Django <= 1.11. **As a result, there is no clean upgrade path for an existing user running the latest releases of Django and django-tenants (1.11 and 1.3.3 respectively).** Any user who wants to either upgrade Django or upgrade django-tenants has to upgrade both packages at the same time. Imagine, if every third-party package did the same, users would have to upgrade them all in one go. There would be too many moving parts and it would be difficult to isolate and debug issues.

I recommend that the latest release supports the latest Django LTS release. This is in fact Django's recommendation as well.

If there is anyway I can be of help, please let me know. If needed, I can create a separate `compat.py` file that handles all compatibility and I volunteer to own and maintain that file.

PR #150, opened 2017-12-07
Support Django 2.0

erikriver:


suriya:
I think we need to come to a consensus on what the default value of the on_delete parameter should be. The value set in the pull request `DO_NOTHING` is not a good choice. It could lead to [errors](https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.DO_NOTHING).

The choices are
 * set `on_delete=CASCADE` since this was the default
 * set `on_delete=PROTECT` if we want to change behavior and offer protection from accidental deletes

I prefer the latter, though I am okay with `CASCADE` as well.

tomturner:
Thank you for this. I am working on Django 2 support and have put some of these changes in already.

PR #157, opened 2018-01-10
Fix TENANT_LIMIT_SET_CALLS implementation

jgnd:
The change introduced in #115 broke the `TENANT_LIMIT_SET_CALLS` because the `DatabaseWrapper` creates a new cursor every time and so the check `self._previous_cursor != cursor` is never false.
This behaviour is described in [PEP 249](https://www.python.org/dev/peps/pep-0249/#connection-methods):
> .cursor()
Return a new Cursor Object using the connection.

and can be seen in the underlying implementation (in [django.db.backends.base](https://github.com/django/django/blob/master/django/db/backends/base/base.py#L231))

I have removed the check and also updated the tests to show the intended behaviour of the `TENANT_LIMIT_SET_CALLS` setting. 

I realise this will probably recreate the issue described in #114 and I tried to figure out the underlying issue but without more information it is hard to diagnose.

### DIFF
--- a/django_tenants/postgresql_backend/base.py
+++ b/django_tenants/postgresql_backend/base.py
@@ -47,7 +47,6 @@ class DatabaseWrapper(original_backend.DatabaseWrapper):
     Adds the capability to manipulate the search_path using set_tenant and set_schema_name
     """
     include_public_schema = True
-    _previous_cursor = None
 
     def __init__(self, *args, **kwargs):
         self.search_path_set = None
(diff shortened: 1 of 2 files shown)


grjones:
It looks like this is only fixed in Django Tenants 2.0.  I think it would be worth backporting this fix to 1.3.x since this is a critical fix, and Django 1.11 will be supported through 2020.

domdinicola:
@tomturner any plan to release a version (actually 2, for django 1.x and 2.x) with this fix?

PR #163, opened 2018-03-12
added makemigrations tip

igncampa:
running ``migrate_schemas --shared`` alone wasn't creating the app.Models (Client and Domain) tables under the public schema in my database. Running ``makemigrations`` before or after migrating schemas solved the issue.


tomturner:
thanks for this

PR #178, opened 2018-07-15
Clone tenant on create

sguermond:
Quick implementation of the functionality for creating new tenants by cloning an existing schema. 

Additional functionality in my branch, wasn't sure if it would be in scope for this PR:
* support for schema names containing uppercase characters or hyphens
* support for `clone_schema` on AWS RDS or other Postgres implementations that do not allow access to the `postgres` superuser

Possible functionality that I haven't written yet:
* managing `TENANT_BASE_SCHEMA` from within django-schema, creating the template and running migrations on it

Still need to add tests, but not entirely sure where to start.


django_tenants/models.py
```diff
@@ -154,7 +154,7 @@ def create_schema(self, check_if_exists=False, sync_schema=True,
                              verbosity=verbosity)
             else:
                 # create the schema
-                cursor.execute('CREATE SCHEMA %s', (self.schema_name,))
+                cursor.execute('CREATE SCHEMA %s' % self.schema_name)
```
kosz85: http://initd.org/psycopg/docs/cursor.html#cursor.execute
This is asking for failure. You should leave processing params to psycopg2, as it map it to proper type using internal mapping system.

sguermond: Found the bug: the branch I originally worked this out on supports schema names with capitalization, so there I'm passing `psycopg2.extensions.AsIs(schema_name)` instead of just `schema_name`. I'd reverted to the old notation here so I could keep the functionality separate. Will fix.

kosz85: Nice one :+1:

tomturner: @sguermond Thank you for hard work. I will merge once you send the latest changes and i have tested them.
@kosz85 Thank you for doing some of the code reviewing.

tomturner: @sguermond did you manage to fix the bug?

tomturner: @sguermond did you manage to fix the bug?

Issue #122, opened 2017-04-06
Does django-tenants support replication ?


thelinuxer:
Hi, 

Having all tenants in a single database is more reason to support replication to avoid bottlenecks. From the code I can't see how possibly you could be handling routing to different databases based on reads and writes. From what I can see the database wrapper is set only for the default connection when connection.set_schema connection.set_tenant is called, am I right ? In that case it's expected to fail to read from the read replica since the schema is only set on the default database.

Are you planning to support this feature in the near future ? We can try to find a solution for it together.

I assume the consumer would have configuration like the following:

A very simple router:
```python
class PrimaryReplicaRouter(object):
    def db_for_read(self, model, **hints):
        """
        Reads go to a replica.
        """
        return 'replica'

    def db_for_write(self, model, **hints):
        """
        Writes always go to primary.
        """
        return 'default'
```

and in the settings.py, the DATABASES configuration:

```python
DATABASES = {
    'default': {
        'ENGINE': 'django_tenants.postgresql_backend',
        'NAME': 'userdb',
        'USER': 'master',
        'PASSWORD': 'password',
        'HOST': 'localhost',
        'PORT': 5432,
        'ATOMIC_REQUESTS': True
    },
    'replica': {
        'ENGINE': 'tenant_backend',
        'NAME': 'userdb',
        'USER': 'master',
        'PASSWORD': 'password',
        'HOST': 'localhost',
        'PORT': 5433,
        'ATOMIC_REQUESTS': True
    },
}

DATABASE_ROUTERS = (
    'django_tenants.routers.TenantSyncRouter',
    'replicationrouter.PrimaryReplicaRouter',
)
```

tomturner:
@thelinuxer Happy to have a PR for this

jcass77:
Should this issue be closed now that #179 has been merged?

mmcclelland1002:
@tomturner @jcass77 Made a PR for this feature: https://github.com/django-tenants/django-tenants/pull/760
Please review whenever you have availability. Thanks! :)

PR #179, opened 2018-07-15
Allow using non-default databases for schemas

sguermond:
Should address replication issues in #122

### DIFF
--- a/django_tenants/management/commands/migrate_schemas.py
+++ b/django_tenants/management/commands/migrate_schemas.py
@@ -1,8 +1,5 @@
-
-from django.db import DEFAULT_DB_ALIAS
-
 from django_tenants.migration_executors import get_executor
-from django_tenants.utils import get_tenant_model, get_public_schema_name, schema_exists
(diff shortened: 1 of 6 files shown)


tomturner:
So does this allow the ability to have different types of schemas?

Also what Documentation is required so the patch?

webapp-dev:
What is the status on this? I see a time soon when my multi-tenant db is full (there are performance limits). ideally I want tenants 1-1000 in db1 1001-2000 in db2 etc etc yet using db1 for the

webapp-dev:
What is the status on this? I see a time soon when my multi-tenant db is full, there are performance limits we are all going to hit. Ideally I want tenants 1-1000 in (db1) 1001-2000 in (db2) etc etc. and at the same time using db1 for the public tenant and for the authentication. Is this possible in django-tenants current form? any progress in this area? Should be achievable when authenticatng on (db1) to get the details of the database to login to for the multi-tenant application side... Any forks attempting this? any anything?

Issue #188, opened 2018-09-12
CURRENT_SCHEMA_TO_SERVER_STATICFILES usage


coler-j:
The docs say:

> IMPORTANT: For all the other things **like collectstatic files,** medias files, the variable CURRENT_SCHEMA_TO_SERVER_STATICFILES it's not necessary, because the schema name is automatically found for these actions.

I have found this to be untrue in development. 

With `CURRENT_SCHEMA_TO_SERVER_STATICFILES = "tenant_b"`, when the command `./manage.py collectstatic_schemas --schema=tenant_a` is run the fileds from `tenant_b` will be collected into the static directory for `tenant_a`.

Any suggestions?

coler-j:
Furthermore, with `DEBUG = True` or `DEBUG = False` (where I can only assume `DEBUG = False` means i am out of my "local development environment" according to the docs), only the static files of the `CURRENT_SCHEMA_TO_SERVER_STATICFILES` schema are used in templates.

coler-j:
`TenantFileSystemFinder` also requires `CURRENT_SCHEMA_TO_SERVER_STATICFILES` to be set or it raises an error... so what should it be set to in non-development environments?

coler-j:
I created my own static file finder for tenants.

Issue #187, opened 2018-09-12
Specialized tenant template: 


coler-j:
I am using Django 2.0.7

I keep getting the error:
`django.core.exceptions.SuspiciousFileOperation: The joined path (C:\....\MyApp\static\tenants) is located outside of the base path component (C:\....\MyApp\static\tenants\sample_tenant)`

When i run the command:
`python manage.py collectstatic_schemas --schema=sample_tenant`

These are my settings, I don't see anywhere where I am varying from the documentation:

```
# Define STATIC_ROOT for the collectstatic command
STATIC_ROOT = join(BASE_DIR, 'static')
STATIC_URL = '/static/'
STATICFILES_STORAGE = 'django_tenants.staticfiles.storage.TenantStaticFilesStorage'
STATICFILES_FINDERS = [
    'django_tenants.staticfiles.finders.TenantFileSystemFinder',  # first
    'django.contrib.staticfiles.finders.FileSystemFinder',
    'django.contrib.staticfiles.finders.AppDirectoriesFinder',
    ]

MULTITENANT_STATICFILES_DIRS = [
    join( BASE_DIR, 'tenants/%s/static'),
]
MULTITENANT_RELATIVE_STATIC_ROOT = "tenants"
CURRENT_SCHEMA_TO_SERVER_STATICFILES = "sample_tenant"
```

_My tenant specialized media files and templates are working fine._

coler-j:
Seems to be completely broken on Django 2.0.7, does it work on lower versions?

coler-j:
I create my own solution.

Issue #158, opened 2018-02-02
Error when I try to use template based on tenants, on Django 2.0.2


moacirmoda:
In django 2.0.2 i have found this error:

```
'FilesystemLoader' object has no attribute 'get_contents'
```

This error occur when I try to use templates based on tenant.

The documentation is deprecated, so I've searched new solutions, until I got this:
```python
TEMPLATES = [
    {
        'BACKEND': 'django.template.backends.django.DjangoTemplates',
        # 'DIRS': [],
        # 'APP_DIRS': True,
        'OPTIONS': {
            'loaders': [
                'django_tenants.template_loaders.FilesystemLoader',
                'django.template.loaders.filesystem.Loader',
                'django.template.loaders.app_directories.Loader',
            ],
            'context_processors': [
                'django.template.context_processors.debug',
                'django.template.context_processors.request',
                'django.contrib.auth.context_processors.auth',
                'django.contrib.messages.context_processors.messages',
            ],
        },
    },
]

MULTITENANT_TEMPLATE_DIRS = [
    'tenant/templates/%s',
]
```

Can anyone help me to find if there is anything wrong with my configurations?

moacirmoda:
Apparently I do some workaround that seems to work. I created a class that based on `FilesystemLoader` inserting `get_contents` method (copied from [here](https://github.com/django/django/blob/2.0.2/django/template/loaders/filesystem.py#L21)), and subscribing method `get_template_sources`, that changes the yield to return instead of a `str` object, a `Origin` object (copied from [here](https://github.com/django/django/blob/2.0.2/django/template/loaders/filesystem.py#L42)).

See the whole code:
```python
from django.conf import settings
from django.core.exceptions import ImproperlyConfigured, SuspiciousFileOperation
from django.db import connection
from django.template import TemplateDoesNotExist, Origin
from django.utils._os import safe_join
from django_tenants.postgresql_backend.base import FakeTenant
from django_tenants.template_loaders import FilesystemLoader as TenantFilesystemLoader


class FilesystemLoader(TenantFilesystemLoader):
    def get_contents(self, origin):
        try:
            with open(origin.name, encoding=self.engine.file_charset) as fp:
                return fp.read()
        except FileNotFoundError:
            raise TemplateDoesNotExist(origin)

    def get_template_sources(self, template_name, template_dirs=None):
        """
        Returns the absolute paths to "template_name", when appended to each
        directory in "template_dirs". Any paths that don't lie inside one of the
[...]

Issue #196, opened 2018-10-31
What is DJANGO<2 in Travis config? 


atodorov:
What is this doing in `.travis.yml` ? 

```
   - $DJANGO="Django<2"
install:
  - pip install -q "$DJANGO"
```

From what I can tell it installs an older version of Django but then, when you install django-tenants via setup.py it picks up the latest version of Django so the above code doesn't make sense.

tomturner:
Fixed

PR #198, opened 2018-11-02
Update template loaders and static file storage for Django > 2.0 

### LINKED ISSUE #197: MULTITENANT_RELATIVE_STATIC_ROOT not working as intended
According to the [documentation](https://github.com/tomturner/django-tenants/blob/12b6a852236363c8478c944e79b1432f5c0f3c46/docs/static.rst#examples), MULTITENANT_RELATIVE_STATIC_ROOT is supposed to default to ``tenant_name``.

This does not appear to be happening, and neither is ``%s``  resolved to the current schema name when used as part of a dynamic path configuration.

The end result is that a ``django.core.exceptions.SuspiciousFileOperation`` exception is raised.

This was also previously reported in #187 though no solution was provided.

jcass77:
It looks like TenantStaticFileStorage has not been updated since Django 1.8, and is no longer compatible with Django > 2.0. 

This PR just rebases the implementation on the standard django.template.loaders package to bring django-tenants back up to parity.

Fixes #197.


jcass77:
I think this PR can be cleaned up further to align the implementation more closely with Django > 2.0.

It is currently a work in progress and I should be able to push the final version in the next few days.

jcass77:
Ok, I think this PR is ready for review. It resolves a large number of long-standing issues related to static file handling.

Apologies for the large commit - most of the diffs are generated by documentation changes and new test cases.

- re-based implementation on Django 2.0 static file finders and storage handlers. Fixes #197. Fixes #187.
- static files defined in the tenant folders will now correctly override any project-level files of the same name. This makes it possible to have tenants selectively change just those files (e.g. images, CSS or JavaScript), that are unique to a particular tenant. 
- ``collectstatic_schemas`` now collects the static files to the correct subdirectory for each tenant.
- Added new template loader. Fixes #158.
- got rid of ``CURRENT_SCHEMA_TO_SERVER_STATICFILES`` hack. Fixes #188.
- added test cases for everything related to static file handling.
- added sensible defaults to make configuration easier.
- rewrote the documentation to simplify configuration steps.
- updated travis django dependency to Django>=2.0,<2.2 to align with setup.py. This would cause the travis build to fail with the new implementation. Also addresses #196.
- updated docker image to use ubuntu:16.04 so that we can get access to Python 3.6 for testing.

phalt:
+1 this will resolve the tests, which do not currently pass when running `setup.py` locally

tomturner:
@jcass77 Thanks for this.

rdraju:
django_tenants.template.loaders.filesystem.Loader get initialized only once. It happens with the first request and directories are set based on that tenant. Same template directories get utilized for all further requests from any tenant.

rdraju:
Not relying on self.dirs and overriding get_dirs is the way to go? Any performance considerations? Can they be mitigated using CachedLoader?

jcass77:
Hmmm...I thought I tested that scenario but you are right @rdraju, the Loader is only initialised once so can see how the same template might be returned for all subsequent tenants.

Have a look at ``TenantFileSystemFinder`` for ideas on doing lazy loading so that the dirs are still only initialised once for each tenant.

selvara80:
Unfortunately, This change breaks  other functionalities.
1)STATICFILES_FINDERS  is only looking for static files from the tenant folder. No file is fetched from the public folder.
2)When testing Media, No module named 'django_tenants.files.storages' .Believe the file name incorrectly got changed to 'storage'

jcass77:
1. As far as I can recall, the standard Django finders will take care of looking up static files in the public folder. You still need to reference it in your ``settings.py`` though, which should include something like the following:
```
STATICFILES_FINDERS = [
    # https://django-tenants.readthedocs.io/en/latest/static.html#static-files-finders-aware-tenant
    "django_tenants.staticfiles.finders.TenantFileSystemFinder",  # Must be first
    "django.contrib.staticfiles.finders.FileSystemFinder",  # <-- include Django standard finder.
    "django.contrib.staticfiles.finders.AppDirectoriesFinder",
]
```
2.``django_tenants.files.storages`` was renamed to ``django_tenants.files.storage`` to align more closely with the Django standard naming convention. I updated all of the internal references but might have underestimated the impact that this could have on existing users with pre-existing configurations. Have you tried updating your settings to use the new module name?

gnalbandian:
> 1. As far as I can recall, the standard Django finders will take care of looking up static files in the public folder. You still need to reference it in your `settings.py` though, which should include something like the following:
> 
> ```
> STATICFILES_FINDERS = [
>     # https://django-tenants.readthedocs.io/en/latest/static.html#static-files-finders-aware-tenant
>     "django_tenants.staticfiles.finders.TenantFileSystemFinder",  # Must be first
>     "django.contrib.staticfiles.finders.FileSystemFinder",  # <-- include Django standard finder.
>     "django.contrib.staticfiles.finders.AppDirectoriesFinder",
> ]
> ```
> 2.`django_tenants.files.storages` was renamed to `django_tenants.files.storage` to align more closely with the Django standard naming convention. I updated all of the internal references but might have underestimated the impact that this could have on existing users with pre-existing configurations. Have you tried updating your settings to use the new module name?

The issue seems to be that the {% static %} is always replaced by the STATIC_ROOT path and the schema name of the tenant appended ( STATIC_ROOT/tenant_schema ) even though there is no specific folder for tenant static files

jcass77:
@gnalbandian can you log a ticket so that we can track an resolve this separately.

If there are no tenant-specific static files for the project, then removing it from the ``STATICFILES_FINDERS`` config section and relying on the Django standard ``django.contrib.staticfiles.finders.FileSystemFinder`` should be all that is required for that specific use case - or am I overlooking something else?

jcass77:
Thanks for the feedback @gnalbandian. Yes I think that is the intended behaviour: each ``STATICFILES_FINDERS`` only looks for static files in the specific folders that it has been designed / configured for.

For ``django.contrib.staticfiles.finders.FileSystemFinder``, that means tenant-specific files.

PR #215, opened 2018-12-23
Start changelog

jcass77:
I was going through the commits since the v2.0.0 release, and thought it might be helpful to start maintaining a changelog?

These entries represent the most significant un-released changes - as far as I could tell.


tomturner:
Thanks for this

PR #223, opened 2019-01-14
create TenantAdminMixin

domdinicola:
Create Mixin to be usable for Tenant Model.
It will help disabling save and delete buttons when you're not in the current or public tenant, preventing Exceptions.


tomturner:
@domdinicola Thanks

domdinicola:
good!

PR #228, opened 2019-01-22
Custom auth backend compatibility for TenantClient

ishtiaque06:
**Rationale**: I'm creating a B2B multitenant app that logs in user using a custom auth backend which uses e-mail as one of the identifiers and authenticates users by parsing `request.META['HTTP_HOST']` to get the relevant schema. However, by doing this, I wasn't able to use `TenantClient`'s `login()` method since my custom auth backend needs a request object but django client's `login` method chain (`login -> authenticate(**credentials)->... ` [as shown here](https://github.com/django/django/blob/master/django/test/client.py#L594)) doesn't pass on a `request` object to my custom auth backend. 

Since I feel like B2B multitenant apps may often use custom auth backends like I did, I felt it necessary to propose this change.

### DIFF
--- a/django_tenants/test/client.py
+++ b/django_tenants/test/client.py
@@ -76,3 +76,19 @@ class TenantClient(Client):
             extra['HTTP_HOST'] = self.tenant.get_primary_domain().domain
 
         return super().delete(path, data, **extra)
+    
+    def login(self, **credentials):
+        # Create a dummy HttpRequest object and add HTTP_HOST
+        from django.http import HttpRequest
+        request = HttpRequest()
+        request.META['HTTP_HOST'] = self.tenant.get_primary_domain().domain
+        
+        # Authenticate using django contrib's authenticate which passes the request on 
+        # to custom backends
+        from django.contrib.auth import authenticate
+        user = authenticate(request, **credentials)
+        if user:
+            super(TenantClient, self)._login(user)
+            return True
+        else:
+            return False


tomturner:
@ishtiaque06 I can't see a problem this going in I will test it and merge in.

Thanks

Tom

lorinkoz:
@ishtiaque06 Question: why not use the provided request factory?

ishtiaque06:
> @ishtiaque06 Question: why not use the provided request factory?

@lorinkoz I'm not sure I understand how I'd go about doing that. Could you point me to the right direction?

ishtiaque06:
> @ishtiaque06 Question: why not use the provided request factory?

Just read the docs. I will try this out to see if it works and keep you updated. Thank you! This is a much neater way of doing things (not up in the "[Advanced Testing](https://docs.djangoproject.com/en/2.0/topics/testing/advanced/)" alley yet).

lorinkoz:
There is a `RequestFactory` for testing purposes at https://github.com/tomturner/django-tenants/blob/master/django_tenants/test/client.py

I think this is the recommended way to create requests for testing.

ishtiaque06:
> There is a `RequestFactory` for testing purposes at https://github.com/tomturner/django-tenants/blob/master/django_tenants/test/client.py
> 
> I think this is the recommended way to create requests for testing.

So I tried out `TenantRequestFactory`. It's pretty nifty. However, when I'm using this, I can't conveniently test various aspects of the `response` that I would be able to access with the TenantClient like `assertTemplateUsed` or `response.context['variable_goes_here']` to see if the right queryset is being sent by the views I'm testing, which for me is the point of unit testing in Django.

ishtiaque06:
> @ishtiaque06 I can't see a problem this going in I will test it and merge in.
> 
> Thanks
> 
> Tom

This is the custom authentication backend I'm trying to use if you want to work with it:

```python
from django.contrib.auth.backends import ModelBackend
from django.contrib.auth import get_user_model
from django_tenants.utils import schema_context


# Custom backend to enable logging in with e-mail.
# https://stackoverflow.com/questions/37332190/django-login-with-email
class EmailAuthBackend(ModelBackend):

    def authenticate(self, request, username=None, password=None, *args):
        if request == None:
            print ("args:", *args)
        # schema = request.build_absolute_uri().split('.')[0].split('/')[-1]
        schema = request.META['HTTP_HOST'].split('.')[0]
        if (schema == 'dibs-test'):
            return None
        with schema_context(schema):
            UserModel = get_user_model()
            try:
                user = UserModel.objects.get(email=username)
            except UserModel.DoesNotExist:
                return None
            else:
                if user.check_password(password):
                    return user
            return None
```

tomturner:
@ishtiaque06 Thanks I have committed this

PR #235, opened 2019-01-28
Moves documentation link (Quality of Life)

coler-j:
Quickly Googling for this library, and then wanting to navigate to the docs is slightly annoying as the link to the docs is hidden halfway through the main readme file (after links to other doc sites, like the postgreSQL docs).

This PR adds an earlier reference to the doc site.


tomturner:
You could have clicked on the buttons at the top which is what I do.

PR #239, opened 2019-02-15
Switch to psycopg2 dependency rather than psycopg2-binary

### LINKED ISSUE #233: Requirements should reference psycopg2 instead of psycopg2-binary
Per the [psycopg2 docs](http://initd.org/psycopg/docs/install.html#binary-install-from-pypi) (emphasis added):

>Note The -binary package is meant for beginners to start playing with Python and PostgreSQL without the need to meet the build requirements. **If you are the maintainer of a publish package depending on psycopg2 you shouldn’t use psycopg2-binary as a module dependency.** For production use you are advised to use the source distribution.

It would be fine to pin `psycopg2<2.8` as recommended in #174, but it would probably also be fine to leave it floating and ignore the warning.

riconnon:
I know this has been discussed either way and I understand the desire to use `psycopg2-binary` during development but this approach is definitely not the answer.

Per the documentation here:
http://initd.org/psycopg/docs/install.html#binary-install-from-pypi

> If you are the maintainer of a publish package depending on psycopg2 you shouldn’t use psycopg2-binary as a module dependency.

Resolves #233

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ setup(
     ],
     install_requires=[
         'Django >= 2.0,<2.2',
-        'psycopg2-binary',
+        'psycopg2',
     ],
     zip_safe=False,
 )


atodorov:
@riconnon can you rebase b/c setup.py changed after Django 2.2 was released.

atodorov:
@tomturner this patch looks sane, do you mind merging after it is rebased ?

tomturner:
@riconnon Thanks

PR #238, opened 2019-02-12
added plan options to support django>=2.2

domdinicola:
you could merge it when 2.2 is official released, but the code should be retrocompatible

PR #250, opened 2019-03-27
Add Django 2.2rc1 to test matrix

atodorov:
mainly b/c this is going to be released soon and I want to check compatibility before we integrate with your library.


atodorov:
Oops, duplicate of #238

PR #251, opened 2019-03-29
Fix create tenant management command

safaariman:
create_tenant management command was not working because of a missing "domain_" prefix. Here is the fix. I also added atomic to "handle" method so that if anything fails nothing will hit to database.


tomturner:
@safaariman Thanks and merged

Issue #249, opened 2019-03-21
TenantFileSystemStorage doesn't work anymore


stygmate:
it's pretty easy to view the problem
just launch:
``` sh
./manage.py tenant_command shell --schema=schema1
```
and in the shell:
``` python
from django.core.files.storage import default_storage
print(default_storage.base_location)
```
the output will be:
```
/media/public
```

stygmate:
@tomturner can you check that ?

atodorov:
In my testing, starting with clean MEDIA_ROOT directory whichever tenant receives the first file upload ends up determining the directory path structure.

So if I first upload to the public schema I get `uploads/tenants/public/...` and everything ends up there.

If I first upload to a schema called `aero` I get `uploads/tenants/aero/...` and then uploads on public end up here too. Will look into the code for and try to make a PR.

stygmate:
@atodorov i'm not behind the code but if i remember they remove path and url method overide of FileSystemStorage in TenantFileSystemStorage and keeped overided code only in __init__
...

jcass77:
Should these changes perhaps also apply to `TenantStaticFilesStorage` as it inherits from `FileSystemStorage` too?

atodorov:
@jcass77 maybe but I don't use that storage nor have the capacity to figure out how not to break existing behavior for people who are using the other storage classes.

stygmate:
@atodorov maybe write a alternative BetterTenantFileSystemStorage class and pull it :) ?

alexandernst:
I can confirm that this also applies to `TenantStaticFilesStorage`. Whatever tenant receives the first request get's to be the default and only for all requests, no matter to which tenant the requests are being directed to.

jcass77:
Linking in related issue #248.

alexandernst:
Here is the MR for this issue (the static files part): https://github.com/tomturner/django-tenants/pull/255

alexandernst:
@atodorov Can you fix your unit tests in #252 so @tomturner can review it, and also review #255 and merge them if they look fine?

atodorov:
> @atodorov Can you fix your unit tests in #252 

No, because I don't know how! 

As I've said in https://github.com/tomturner/django-tenants/pull/252#issuecomment-478183296 using the standard executor everything works and I've also verified that manually but I have no idea why the multiprocessing executor fails and I am asking for help to debug or hints how to figure out what the problem is.

[...]

PR #252, opened 2019-03-29
Update test config and TenantFileSystemStorage for #249

atodorov:
First step, this got docker running for me and tests are PASS.

I believe the current file storage class and its tests are wrong and thus causing #249 so this is what I am going to investigate next (and possibly push to the same PR if you don't merge before that).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,8 +5,5 @@ RUN apt-get update && apt-get -y upgrade
 
 RUN apt-get -y install postgresql libpq-dev postgresql-client postgresql-client-common python3-pip git netcat
 
+VOLUME /code
 WORKDIR /code
-
-ADD . /code
-
-RUN python3 setup.py develop
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ MAINTAINER Przemek Kamiński <cgenie@gmail.com>
 
 RUN apt-get update && apt-get -y upgrade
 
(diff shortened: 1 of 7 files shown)


atodorov:
@tomturner - with all of these changes I'm able to execute the tests locally and they pass, files are saved under different directories based on the schema_name.

The failure comes from the multiprocessing executor telling me that migration already exists. From what I can tell a tenant already exists in the DB but for some reason we try to create it again. I need a hint on this one.

Another question: do you mind including this and other open PRs in a release after Django 2.2 comes out?

tomturner:
@atodorov regrading PR going in before the next release it depend if it ready in time. I will only merge if passes Travis the test I will look however I am quite busy so other peoples help would be good

atodorov:
> @atodorov regrading PR going in before the next release it depend if it ready in time. I will only merge if passes Travis the test I will look however I am quite busy so other peoples help would be good

+1 here, I'am also quite busy but need some hints how to debug the failure.

It seems to be only related to the multiprocessing executor. Not sure why it is failing. Is it by chance that it is trying to re-use the same database/migrations which already exist ? I'm not even sure what this multiprocessing executor is and why we need it in the tests.

alexandernst:
I'm trying to debug the multiprocessing problem in my branch, but it just seems like there is some major wtf going on. I see migrations being ran multiple times per tenant and per test. 

I'll keep this updated.

atodorov:
Superseded by #255 so closing.

PR #255, opened 2019-04-02
Fix #249 (static files) (supersedes #252)

### LINKED ISSUE #249: TenantFileSystemStorage doesn't work anymore

alexandernst:
This handles the bug described in #249 , the part related to static files.


alexandernst:
@atodorov @tomturner Unit tests fixed ^. The main problem with the unit tests per-se is that atomic commits (as per TestCase vs TransactionTestCase for the base of the unit tests) are used everywhere, both when wrapping the classes and when wrapping each unit test, which makes debugging extremely hard. 

This particular bug was caused by TenantTestCase creating things that the test itself is already creating.

alexandernst:
Oh, also, this MR supersedes #252 , so don't merge #252 .

atodorov:
@tomturner what do you think about this change ? It is rather big but most of the things are support around running the test suite with docker.

PR #256, opened 2019-04-07
Fix #248

### LINKED ISSUE #248: Static Files Won't Serve
Following the docs here: https://django-tenants.readthedocs.io/en/latest/files.html and can't get static files to serve

I run `python manage.py collectstatic_schemas --schema=tenant1`

Static files gets copied to `my_proj_path/staticfiles/tenant1` but logging into the admin via the frontend still yields 404 with css/js files. within the tenant1 folder I see the normal django `admin` folder with the static resources.

Chrome is looking for these files at (example: `http://tenant.localhost:8000/static/tenant1/admin/css/base.css`) which looks correct and the same as the folder structure where collectstatic_schemas is writing to.

I believe I am following the docs very closely but can't get static files to serve. Any help would be appreciated

The relevant parts of my settings files are:

    # This gets added to `INSTALLED_APPS` as described with the docs
    SHARED_APPS = (
        'django_tenants',
        'playground.customers',
        'django.contrib.staticfiles',
    )

    STATIC_URL = '/static/'
    STATIC_ROOT = os.path.join(BASE_DIR, 'staticfiles')

    # imported from: `from django.conf.global_settings import STATICFILES_FINDERS`
    STATICFILES_FINDERS.insert(0, "django_tenants.staticfiles.finders.TenantFileSystemFinder")

    MULTITENANT_STATICFILES_DIRS = [
        os.path.join(BASE_DIR, "tenants/%s/static"),
    ]

    STATICFILES_STORAGE = "django_tenants.staticfiles.storage.TenantStaticFilesStorage"

    MULTITENANT_RELATIVE_STATIC_ROOT = ""  # (default: create sub-directory for each tenant)

alexandernst:
Detailed explanation in the diff

### DIFF
--- a/django_tenants/staticfiles/finders.py
+++ b/django_tenants/staticfiles/finders.py
@@ -1,4 +1,5 @@
-from django.contrib.staticfiles.finders import FileSystemFinder
+import os
+from django.contrib.staticfiles.finders import AppDirectoriesFinder, FileSystemFinder
 from django.conf import settings
 from django.core.checks import Error
 from collections import OrderedDict
@@ -8,6 +9,38 @@ from django.db import connection
 from django_tenants import utils
 from django_tenants.staticfiles.storage import TenantStaticFilesStorage
 
+class TenantAppDirectoriesFinder(AppDirectoriesFinder):
+    """
+    This is a replacement for the AppDirectoriesFinder. The only difference between
+    this and the standard AppDirectoriesFinder implementation is that this one will
+    understand how tenant-aware file handling works and where to search for your
+    files.
+
+    In-detail explanation:
+
+    Let's say that we have a Tenant called "Foo" (foo.com).
+    When we try to access, let's say, foo.com/admin, the browser will issue a request
+    to (among others) foo.com/static/foo/admin/css/base.css , instead of the
+    usual request if we didn't use django-tenants (foo.com/static/admin/css/base.css).
+
+    See the "foo" right before "admin"?
+
+    In order for this to work as expected we must remove the "foo", as we have
+    already added the path "tenants/%s/static" to the paths that Django will
+    look.
+
+    If we didn't remove the "foo" from the path, Django would look in
+    /code/tenants/foo/static/foo/admin/css/base.css , which is wrong.
+
+    Note: Keep in mind that this code won't run at all when behind uWSGI/NGINX/whatever
+    web server you're using. Django won't even know where your static files are at!
+    """
+    def find_in_app(self, app, path):
+        """
+        Find a requested static file in an app's static locations.
+        """
+        path = os.path.sep.join(path.split(os.path.sep)[1:])
+        return super().find_in_app(app, path)
 
 class TenantFileSystemFinder(FileSystemFinder):
     """
@@ -26,6 +59,19 @@ class TenantFileSystemFinder(FileSystemFinder):
         # the first time they are needed.
         self._locations = {}
         self._storages = {}
+
+    def find_location(self, root, path, prefix=None):
+        """
+        Find a requested static file in a location and return the found
+        absolute path (or ``None`` if no match).
+        """
+        # This will remove the name of the tenant from the path. This will
+        # allow is to correctly match the path when using Djang's built-in
+        # server.
+        # Note that this doesn't matter at all when running in production
+        # mode as Django won't be handling the static files.
+        path = os.path.sep.join(path.split(os.path.sep)[1:])
+        return super().find_location(root, path, prefix)
 
     @property
     def locations(self):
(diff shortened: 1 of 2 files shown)


alexandernst:
@tomturner This fixes #248 . What do you think?

PS: I'll create another patch for the multi-threading issue I found.

tomturner:
@alexandernst I couldn't get this work in fact it made the it not work at all. I think the best thing to do is do a Skype call and share screen so you can show me the problem or you can create a video/example app of the problem. I won't merge until I have the same problem and can recreate it.

alexandernst:
Hi! I'll provide you with a detailed explanation and a demo so you can the issue by yourself ;) 
I just need a couple of days, I'm in the middle of a huge refactor in another project and I don't have any brain cycles left.

alexandernst:
@tomturner Sorry for taking so long. Here you go a vainilla django project with django-tenants and a README.md explaining how to reproduce the bug.

I also setup a docker compose file so you can test it easily.

[example.zip](https://github.com/tomturner/django-tenants/files/3086175/example.zip)

docs/files.rst
```diff
@@ -59,7 +59,7 @@ Start by inserting django-tenants' ``django_tenants.staticfiles.finders.TenantFi
     STATICFILES_FINDERS = [
         "django_tenants.staticfiles.finders.TenantFileSystemFinder",  # Must be first
         "django.contrib.staticfiles.finders.FileSystemFinder",
-        "django.contrib.staticfiles.finders.AppDirectoriesFinder",
+        "django_tenants.staticfiles.finders.TenantAppDirectoriesFinder", # This replaces Django's AppDirectoriesFinder
```
jcass77: Instead of removing the standard ``AppDirectoriesFinder``, consider just inserting ``TenantAppDirectoriesFinder`` above it.

This way, static file lookups can fall back to using the Django standard finders if there are no tenant-specific files to serve.

django_tenants/staticfiles/finders.py
```diff
@@ -8,6 +9,39 @@
...
+    understand how tenant-aware file handling works and where to search for your
+    files.
+
+    In-detail explanation:
+
+    Let's say that we have a Tenant called "Foo" (foo.com).
+    When we try to access, let's say, foo.com/admin, the browser will issue a request
+    to (among others) foo.com/static/foo/admin/css/base.css , instead of the
+    usual request if we didn't use django-tenants (foo.com/static/admin/css/base.css).
+
```
jcass77: My suggestion as part of #265 and #266 is that django-tenant's ``TenantStaticFilesStorage `` should not be injecting the ``schema_name`` in static URLs (thereby breaking any standard downstream finders).

Instead, if the tenant-aware finders fail to locate a static file, then the standard finders should serve as fallback.

django_tenants/staticfiles/finders.py
```diff
@@ -8,6 +9,39 @@
...
+    In order for this to work as expected we must remove the "foo", as we have
+    already added the path "tenants/%s/static" to the paths that Django will
+    look.
+
+    If we didn't remove the "foo" from the path, Django would look in
+    /code/tenants/foo/static/foo/admin/css/base.css , which is wrong.
+
+    Note: Keep in mind that this code won't run at all when behind uWSGI/NGINX/whatever
+    web server you're using. Django won't even know where your static files are at!
+    """
```
jcass77: Yes. Luckily, Django only saves the first file that it finds to ``STATIC_ROOT`` when we run ``collectstatic_schemas``.

If everything goes as planned then our tenant-specific static files will be found first (since these finders are configured higher up in ``STATICFILES_FINDERS``), and the tenant-specific versions of the static files will also be the ones served up by the web server.

docs/files.rst
alexandernst: The `TenantAppDirectoriesFinder` is just a wrapper on top of the `AppDirectoriesFinder`, so it doesn't really matter. Even if I leave the `AppDirectoriesFinder`, it won't trigger ever, as the `TenantAppDirectoriesFinder` will find whatever `AppDirectoriesFinder` is supposed to find.

django_tenants/staticfiles/finders.py
alexandernst: Let me try this, but I think there will be problems once the project is deployed and server by uWSGI.

tomturner:
@alexandernst thanks for the demo I will be looking at this shortly

alexandernst:
@tomturner ping me when you're on it and maybe I can help you with the revision process

tomturner:
@alexandernst I haven't forgotten this. I should be able to get this done within the week just quite busy at the mo.

tomturner:
@alexandernst Thanks

PR #261, opened 2019-04-15
add templaes to package

domdinicola:
templates are not included in the package (my bad!!)
bonus: allow customization of app colors though extrastyle block


tomturner:
@domdinicola Thanks

tomturner:
@domdinicola released on PyPi

PR #262, opened 2019-04-15
include_package_data for templates

domdinicola:
https://python-packaging.readthedocs.io/en/latest/non-code-files.html


domdinicola:
@tomturner thanks! Now should be fine :+1:

PR #264, opened 2019-04-19
Document migrating single-tenant to multi-tenant. Fixes #247

### LINKED ISSUE #247: How to move data from existing (single schema) database to multitennant
Is there anyone that have tips and tricks on how to migrate the data?
i've a database with all the data, i would like to use this package but i've also to move the data. most of the models have a relation owner=user, but i've no clear idea on how to write a migration that takes from the old db and moves to the table in the correct schema.

quite sure i'm not the first one that wants to add this to an existing project, anyone that can help?

atodorov:
these instructions work for us and are failry portable between DB backends, however they may not always work hence the warning!


tomturner:
@atodorov

PR #266, opened 2019-04-21
Fix finding of local tenant-specific application static files.

### LINKED ISSUE #265: TenantStaticFilesStorage is incompatible with FileSystemFinder.
The URLs generated by ``TenantStaticFilesStorage`` are not compatible with the standard Django ``FileSystemFinder``.

This causes a range of issues as soon as users swop out the standard ``StaticFilesStorage`` via the recommended ``STATICFILES_STORAGE = "django_tenants.staticfiles.storage.TenantStaticFilesStorage"`` configuration setting.

For example, if tenant 'public' is active, then [``TenantStaticFileStorage.base_url``](https://github.com/tomturner/django-tenants/blob/97e2240a874499a631f9f6d205c425e331cff9d2/django_tenants/staticfiles/storage.py#L32-L33) will append the ``schema_name`` to the URL to produce ``/static/public/``.

This behaviour has some unintended consequences:

1. It is likely that most downstream finders will be based on Django's ``FileSystemFinder``. Since the standard finder is not tenant-aware, and does not know how to interpret tenant-specific URLs, all of those finders will fail to find any static files (tenant-specific or otherwise), essentially rendering them useless.

    This is probably also why @gradedcatfood got his deployment to work as part of #248 by adding the ``compressor.finders.CompressorFinder`` (since that finder ignores ``STATICFILES_STORAGE`` and uses a hard-coded ``CompressorFileStorage`` for serving up compressed files). 

2. A very useful aspect of how django-tenant's static file handling is intended to work is that users are supposed to be able to selectively override static files with tenant-specific versions, without the framework insisting that each tenant **has** to provide its own version of **every** static file. This is explained in the [docs](https://django-tenants.readthedocs.io/en/latest/files.html#configuring-the-static-file-finders):

>...by adding TenantFileSystemFinder at the top, we ensure that Django will look for the tenant-specific files first, before reverting to the standard search path. This makes it possible for tenants to override any static files (e.g. stylesheets or javascript files) that are specific to that tenant, and use the standard static files for the rest.

An easy way to see this problem manifest in the most recent build is to delete any static files that have been collected into the ``STATIC_ROOT`` folder (usually 'staticfiles') and then try to access the django-admin pages. In this use case, ``TenantStaticFilesStorage`` fails to find any static files in ``STATIC_ROOT``, and all of the fallback finders configured in ``STATICFILES_FINDERS`` choke on the tenant-aware URL:

A-Trade_Backend

jcass77:
Apologies for the large PR; it actually just contains one fix for #265 with some refactoring and a lot of additional tests.

I took particular care to ensure that the tests introduced as part of #249 still works, so hopefully there won't be any regressions on how ``media`` related files are handled.

- Re-align 'staticfiles' implementation with Django standard.

- Refactor tests and increase test coverage.

- Update ``dts_test_project`` to serve as example of a project structure using django-tenant's static files features.

- Update documentation.

- Fixes #265.


jcass77:
Heads-up, I've pushed some more commits to address some valid comments raised by @alexandernst in related PR https://github.com/tomturner/django-tenants/pull/256#discussion_r277164293.

To summarise, this is what we are aiming for:

- during local development, and when collecting static files using ``collectstatic_schemas``, we need to retain compatibility with all of the vanilla Django finders. This ensures that the standard finders can be used as a fallback if no tenant-specific versions of a file has been created or is necessary.

- in staging / production, where we expect files to be served by an external HTTP server like nginx, we need to re-write the URLs to include the tenant details so that the HTTP server will look for the static files in the proper subdirectories of ``STATIC_ROOT`` (usually `/staticfiles/schema_name/`).

The only sensible way I can think of achieving this is to introduce a new configuration setting, ``REWRITE_STATIC_URLS``, that django-tenants can use to know when URL's will be served by Django using the various finders (``REWRITE_STATIC_URLS = False``), and when files have been collected to be served by an external HTTP server (``REWRITE_STATIC_URLS = True``).

alexandernst:
@jcass77 does this supersedes #256 ?

tomturner:
@jcass77 and @alexandernst I was going to as 

> @jcass77 does this supersedes #256 ?

I was going to ask the same question

jcass77:
No, it does not supersede #256.

If there is a need for a tenant-aware ``AppDirectoriesFinder``, (i.e. one that looks for static files in something like ``/app_name/tenants/`` then that PR is still relevant.

It might be easier to get this in first though, and then revisit the approach for implementing a ``TenantAppDirectoriesFinder``.

My gut feel is that we should be aiming for something that can be used in addition to the Django standard apps directories finder, rather than building a replacement.

jcass77:
I was a little optimistic in hoping that the tests for #249 would be enough to ensure backwards compatibility with that PR.

It turns out that we need to add the tenant's `schema_name` when building the MEDIA_ROOT URL as well.

I've pushed some small changes to align the implementation of `TenantFileSystemStorage` with that of `TenantStaticFilesStorage` - media files should now be server properly for each tenant.

tomturner:
@jcass77 I need a little bit of time to test before I merge

jcass77:
I've rebased this on the 'master' branch to make reviewing / testing / merging easier.

tomturner:
Thanks I will review

jcass77:
@tomturner, it looks like the test cases for this PR weren't actually being run.

I've pushed a new commit to ensure that the tests are included in `run_tests.sh` (also see #300, which I think might be a better approach overall).

The tests also uncovered a handful of regressions that have subsequently been fixed.

Celant:
Also using this fix branch, and not seen any issues with it!

tomturner:
@jcass77 Thanks

PR #268, opened 2019-04-22
test with proper matrix

domdinicola:


domdinicola:
this PR tests with Django 2.0, 2.1 and 2.2 (before testing was done only on lastest - 2.2)

tomturner:
I will be removing support for django 2.0 as it no longer supported by django

PR #295, opened 2019-07-25
Update files.rst

johnfraney:
Fixes a couple typos, the most important of which is the description of the template loaders that mentioned `django` instead of `django_tenants`.


tomturner:
@johnfraney thanks

PR #298, opened 2019-08-09
add test domain to allowed hosts for FastTenantTestCase

### LINKED ISSUE #297: FastTenantTestCase does not find any tables
Hi there,

I've been using django-tenants happily for a while now, and am now trying to make my unit tests run faster. I've been using the normal TenantTestCase previously and everything worked well (albeit slow). Now, with FastTenantTestCase I'm constantly getting errors that my tables are not found for every database access. (e.g. psycopg2.errors.UndefinedTable: relation "training_lenauser" does not exist).

This even happens if I run only one single test function, where the database is created beforehand.

To make sure the tables really get created I tried running manage.py test with the -k flag to keep the whole database. I checked with psql on the command line, everything is there.
I checked the search path, and it is correctly set to "test, public".

I also checked that django.db.connection.tenant is correct, which it also is.

There are no other code changes than the changed inheritance to FastTenantTestCase. Are there any additional changes I need to make when switching to fast?

Do you have any pointers as to why this could happen? I'm totally out of ideas at this point...

bernhardmiller:
this fixes #297

### DIFF
--- a/django_tenants/test/cases.py
+++ b/django_tenants/test/cases.py
@@ -124,6 +124,7 @@ class FastTenantTestCase(TenantTestCase):
 
     @classmethod
     def setUpClass(cls):
+        cls.add_allowed_test_domain()
         tenant_model = get_tenant_model()
 
         test_schema_name = cls.get_test_schema_name()


tomturner:
@bernhardmiller thanks

PR #300, opened 2019-08-13
More robust test discovery and better test encapsulation

jcass77:
`run_tests.sh` currently only runs tests that form part of the `tests` package, instead of running all available tests in the `tests` folder.

- The package-based approach is problematic since, if a developer forgets to update the relevant `__init__.py` files to include new test modules, the newly added tests will not be included when `manage.py` does its test discovery. `test_settings.py` and `test_fast_tenants.py` are good examples of this.

- This PR switches to a path-based approach so that any `test_*.py` files in the `tests` directory will automatically be included in the test run.

This PR also suggests an approach for providing better encapsulation for tests that are based on `FastTenantTestCase` (which, by design, do not clean up tenants and domains that were created as part of their teardown routines).

- `TenantTestCase` and `FastTenantTestCase` both currently use the 'test' schema and 'tenant.test.com' domain.

- If the test suite contains a combination of these two types of test cases, then `TenantTestCase.setUpClass()` will fail with `Key (schema_name)=(test) already exists` when it tries to create the test tenant (since any previously run `FastTenantTestCase`s would have the side effect of leaving its test tenant and domain in the database).

- A better approach might be to use separate schemas and domains for `TenantTestCase` and `FastTenantTestCase` (e.g. 'fast_test' and 'tenant.fast_test.com') to avoid any conflicts. This will allow `TenantTestCase`s to continue working as expected while still providing support for quick setup using `FastTenantTestCase` where required.


tomturner:
@jcass77 Thanks I will have look at this PR

tomturner:
@jcass77 Any chance of fixing the confict

jcass77:
@tomturner, I have resolved the merge conflicts.

There are two test cases that fail after merging in #256, which I don't think is compatible with #266 - see https://github.com/tomturner/django-tenants/pull/266#issuecomment-485770335.

jcass77:
@tomturner, I've rebased this PR on master: all tests are green.

Issue #303, opened 2019-09-10
base_url is giving a filesystem path instead of url


theiviaxx:
https://github.com/tomturner/django-tenants/blob/7f2c42a3790809e40a7bf42d2f364022f0fec496/django_tenants/files/storage.py#L27

if your `MULTITENANT_RELATIVE_MEDIA_ROOT` is set to an absolute path `/mnt/foo/static` then your urls end up like `/mnt/foo/static/path/to/file.jpg`.  

Would it be better to add `MULTITENANT_RELATIVE_MEDIA_URL` or just use `MEDIA_URL`?

PR #305, opened 2019-09-13
base_url return url with tenant

theiviaxx:
Addresses #303


jcass77:
A lot of the static file handling has changed in #266, which resolves several issues and should ideally be merged ahead of further incremental improvements.

I haven't tested it but suspect that one of those changes, https://github.com/tomturner/django-tenants/pull/266/files#diff-1e5d2611255c44572915a0092dbeed0cR25, should go some way towards addressing #303 as well.

theiviaxx:
The call to `relative_media_url` looks like it could return invalid urls as well:

    MULTITENANT_RELATIVE_MEDIA_ROOT = '/mnt/foo'
    MEDIA_URL = '/foo/static/'

This would yield a url `/foo/static/mnt/foo/my_tenant/image.jpg` which probably isn't what you want.  The `MULTITENANT_RELATIVE_MEDIA_ROOT` should have no bearing on the url, unless i misread the code

PR #307, opened 2019-09-24
Fix #302

### LINKED ISSUE #302: Invalid string used for the schema name. When using only numbers
Before I just make a change I wanted to reach out.

I wanted to create a schema name with just number ie 123 however it doesn't work due to a regular expression 
`SQL_IDENTIFIER_RE = re.compile(r'^[_a-zA-Z][_a-zA-Z0-9]{,62}$')`

I have tried to create the schema manually in PgAdmin and PostgreSQL allows it! 

Does any mind if I change that regular expression? or do people many we to make it a setting?

Tberdy:
Fix a crash (ProgrammingError) when attempting to create a schema named like a SQL keyword (select, where, ...)
Allow to create a tenant named with only numbers (123).

### DIFF
--- a/django_tenants/models.py
+++ b/django_tenants/models.py
@@ -137,7 +137,7 @@ class TenantMixin(models.Model):
         if has_schema and schema_exists(self.schema_name) and (self.auto_drop_schema or force_drop):
             self.pre_drop()
             cursor = connection.cursor()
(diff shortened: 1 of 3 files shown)


tomturner:
@Tberdy Thanks for this

PR #311, opened 2019-10-21
Update dependency psycopg2

gilsonbp:
Changing dependency to `psycopg2-binary`. With it is not necessary to install Postgresql on the host


tomturner:
Please look at https://github.com/tomturner/django-tenants/issues/233  This is the reason it got switched back.

tomturner:
I think the best think to do is remove it out of the setting file then people can do they like

tomturner:
Got rid of the requirement so people can choice what to install

PR #324, opened 2019-12-06
Addresses #303

theiviaxx:
Sorry, i botches the previous PR.

This change doesn't use the relative media root for the base urls.  If you use the media root, you might be using filesystem paths which don't relaly belong in the urls.  For example:

```
MULTITENANT_RELATIVE_MEDIA_ROOT = '/mnt/foo'
MEDIA_URL = '/foo/static/'
```
This would yield a url `/foo/static/mnt/foo/my_tenant/image.jpg` which probably isn't what you want. The `MULTITENANT_RELATIVE_MEDIA_ROOT` should have no bearing on the url, unless i misread the code

### DIFF
--- a/django_tenants/files/storage.py
+++ b/django_tenants/files/storage.py
@@ -53,7 +53,8 @@ class TenantFileSystemStorage(FileSystemStorage):
 
     @property
     def base_url(self):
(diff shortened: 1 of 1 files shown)


theiviaxx:
nvm, i need to learn to read docs and variable names

PR #337, opened 2020-01-20
Use field.attname instead of field.name in create_tenant.py

atodorov:
allows specifying integer values for ForeignKey fields.

For example consider the following TENANT_MODEL:

```
class Tenant(TenantMixin):
    ...
    owner = models.ForeignKey(settings.AUTH_USER_MODEL, ...)
```

Without this change create_tenant will ask for --owner and then error out because objects.create() expects a `User` instance.

With this change create_tenant will ask for --owner_id and it will accept an integer value instead.

@tomturner the only difference here is when you have ForeignKey fields. `field.attname` is the same as `field.name` for regular fields.

If you can merge this and push a new version to PyPI that would be greatly appreciated.

### DIFF
--- a/django_tenants/management/commands/create_tenant.py
+++ b/django_tenants/management/commands/create_tenant.py
@@ -23,12 +23,13 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
 
         for field in self.tenant_fields:
-            parser.add_argument('--%s' % field.name,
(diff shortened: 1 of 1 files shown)


atodorov:
@tomturner ping, any chance you could review and hopefully release this change ?

tomturner:
I will try and release later today?

Issue #335, opened 2020-01-10
Python 3.8 and parallel migration sometimes fails the tests


tomturner:
If you look at travis you will see the build is sometimes failing with Python 3.8. I have looked into this and it to with the parallel migration (multiprocessing). Help fixing this problem would be helpful?

Issue #183, opened 2018-08-21
Tenant selection based on url prefix


ghost:
hello, I created a middleware to get the tenant from the URL (eg.example.com/tenant1/). But url reversing for admin app is not working. How to fix the issue with url reversing in Django while using different tenant selection option.

kosz85:
Using path prefixes is tricky, you may need to replace django url reverse function with your own replacement. If this is a problem you have. As you just said "not working" which is the worst description of problem ever.

tomturner:
I would make all tenants go to a url such as http://example.com/r/tenant1/ then your middleware wouldn’t Clash with the django admin urls

kosz85:
Yeah, that's true I use `/tnt/<tenant>/`, same reasons.

coler-j:
Anyone care to share their tenant selection middleware from URL prefix?

fsladkey:
I'm currently encountering the same problem, and any advice would be appreciated. I've tried to outline some approaches I've tried to give more context.

I am successfully setting the current tenant in the middleware, but  by setting my admin path with `path("/tnt/<tenant>/admin", admin.site.urls)` my admin breaks with the error "{View name} received an unexpected keyword argument 'tenant'". Of course, I don't care about the `tenant` arg here, since the tenant is already set via the middleware, but I understand why the view is complaining. :)

I thought I might be able to create a tenant specific URLconf on the fly similar to the way we set the public, by setting `request.urlconf = settings.ROOT_URLCONF + [path(f"/tnt/{tenant.schema_name}/admin", admin.site.urls)]` but that doesn't work because `request.urlconf`is the path to the settings rather than an array of `urlpatterns`.

I can just manually create an admin url per tenant, something like
```python
urlpatterns = [
   ...
] + [path(f"/tnt/{tenant.schema_name}/admin", admin.site.urls) for tnt in Tenant.objects.all()]
```

However, the trouble there is that if you add a new tenant, the admin URLs don't exist until the application is restarted. As an experiment, I tried reload the URLconf per request, and while that works I don't think it's a reasonable approach to solving the problem.

I'm pretty new to django in general, so I may be missing an obvious solution here. Anything to point me in the right direction would be appreciated since I'm quite confident it's possible to do this in a reasonable way. Also, thanks for your work on `django-tenants`! It's been working out great so far, only running into trouble now that I'm trying to do something non-standard.

lorinkoz:
[...]

Issue #28, opened 2015-07-27
Feature-Request: Change tenant-domain from subdomain to url-parameter?


thebarty:
Hi guys,

I am just thinking: One of the big benefits of this app is the added security layer via database-schemas.
So users are probably quite security concerned and will also invest in an ssl-certificate.

Which is quite expensive for a startup... I think unlimited ssl-subdomains-certificates are around 500$ per year?

**Which brings me to the point:**
How easy do you think would it be to implement a switch "DOMAINTYPE = SUBDOMAIN/URL-PARAMETER"?

So that instead of 

```
tenant.domain.com 
```

we would see 

```
www.domain.com/tenant/*?
```

Jafula:
I use these for US$200 a year.

https://www.rapidssl.com/buy-ssl/wildcard-ssl-certificate/index.html

They are owned by GeoTrust (now part of Symantec).

kosz85:
It's easy, I have such middleware in my projects that supports both methods, thought we mostly use `/tnt/<tenant_name>/` instead of domain. It's just `process_response` part that is selecting current tenant, and probably adding some value to request or some global settings like `settings.URL_PREFIX`.
 It's a bit more painful if you use other tools that do smth with url like django-rest-framework, as you need to hack into `rest_framework.reverse.reverse` and `rest_framework.reverse.reverse_lazy`, to always provide your current `URL_PREFIX`, but after you have it, it allows you to have lang in your paths (`/tnt/test/en/`), and any other feature related to prefixing url path.

For free certs check https://startssl.com.

tomturner:
@kosz85 Any chance in sharing such middle ware that allows this enhancement? I also know it would be easy however if you done it no point in reinventing the wheel again.

kosz85:
Can't do that in easy way :) I have there much more features included, not all compatible with django-tenants. I have schema names in form of hash to fit in 64 chars and avoid conflicts, so my `URL_PREFIX` is based on separate name on Tenant table, I also have domains on Tenant model, like it was in old solution.
I'm already on django 1.8 and I'm using django-tenants, so I will be slowly porting my features to fit better in to your lib, but it's slow process. I can write such middleware in my spare time, but I don't have it to much ;)

lorinkoz:
[...]

PR #370, opened 2020-04-18
Subfolder routing

lorinkoz:
Coarse proof of concept of the mechanism I am using in https://github.com/lorinkoz/django-pgschemas

Something like this could become the solution for issues like #183, #28

Some comments:
- Here I'm using the `schema_name` as identifier for subfolder matching. In `pgschemas` I extended the domain model to become a pair of `(domain, folder)`, so that fine grained routing could be achieved.
- In `pgschemas` I have reports from a user that this method for dynamic urlconf construction fails when using certain 3rd party libraries that do custom handling of URLs (this would be up to testing on real life projects that might contain those). Another variant with less dark magic would be something I saw on #28 in a code example from @owais, the drawback of that solution would be that a new module will be created on the fly for each tenant, but not sure if that would even be a problem per-se.
- In `pgschemas` I had to use `django.urls.clear_url_caches` in order to refresh from the previous matched subfolder tenant, ~maybe it's not needed here because this project is using `set_urlconf`~(I will add this here soon). To be aware, the symptom of this problem is that after successfully visiting tenant A through subfolders, reversed urls for tenant B are returned with the tenant A instead. (see https://github.com/lorinkoz/django-pgschemas/blob/master/django_pgschemas/middleware.py#L57)
- Tests are failing on Python 3.8 (#335)
- If this proof of concept is okay, I can work in adding tests.

### DIFF
--- a/django_tenants/middleware/main.py
+++ b/django_tenants/middleware/main.py
@@ -4,7 +4,8 @@ from django.http import Http404
 from django.urls import set_urlconf
 from django.utils.deprecation import MiddlewareMixin
 
-from django_tenants.utils import remove_www, get_public_schema_name, get_tenant_domain_model
+from django_tenants.utils import remove_www, get_public_schema_name, get_tenant_model, get_tenant_domain_model
+from django_tenants.urlresolvers import subfolder_matcher, subfolder_dynamic_urlconf
 
 
 class TenantMainMiddleware(MiddlewareMixin):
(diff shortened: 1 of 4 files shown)


lorinkoz:
The build for Python 3.8 is failing, no idea why.

tomturner:
Thank you I will have a look. I started trying to implement this last week in a branch https://github.com/tomturner/django-tenants/tree/same_domain
I used the domain as the subfolder name.
the problem I came across was the fact that the url reverse didn't work.

lorinkoz:
I am using the exact same logic that `i18n_patterns` uses to prepend all existing URLs with a languange prefix, in this case, with a tenant prefix. This makes reversing work automatically, reusing the same "magic" Django ships already for internationalization. The major difference here is that when you use `i18n_patterns` you usually put that in your `urlconf` yourself (the user, conscious decision), whereas in this case, we're trying to do some dark magic to construct a module on the fly where the `urlconf` of the user is automatically prepended.

The problem I've found with that, is that in order to construct the module on the fly, I have to execute the original `urlconf` module. When you're using certain 3rd party apps that do URL management on their own, executing the `urlconf` module from a middleware is problematic. If we found a way to do this without executing the `urlconf` module, that would be great.

The other option would be to ask the user to create a separate module where they do the `tenant_patterns` themselves, and use a convention through settings to decide which urlconf module to use depending on the tenant routing. I personally don't like this at all, that's why I consider finding a way to do this part automatically is the ace move for this package.

lorinkoz:
From my hiccups with `pgschemas`, there is another philosphic problem around the corner. Tenants and domains have a one to many relationship. When the tenant is decided from an incoming request, it's plain easy, because we traverse the relationship from the many to one side.

Now, if you ever need to reverse a URL outside of the request/response cycle, the problem becomes getting a proper domain from the current tenant. When you don't have subfolder routing, that's no problem at all, because Django doens't include the host as part of its reversing mechanism. Now, as soon as you have subfolders, you have to decide whether to use subfolder reversing or subdomain reversing.

The challenge is not in the code, it's more in the philosophic side. Suppose you are inside a management command or inside a celery task, and all that you have is a tenant: what API do you give the programmer to decide if they want subfolder or subdomain routing for any subsequent call of `reverse`?

The `reverse` method allows for passing a `urlconf`, however, on some of this scenarios, the programmer might not have access to every occurrence of `reverse` that is coming down the code. In this case, some sort of "domain setting" or "routing scheme setting" should be performed as to guarantee that all subsequent URL reversing will be consistent with the programmer's decision.

tomturner:
@lorinkoz thanks for that

tomturner:
@lorinkoz Your commit is far better than what I did! The only think i was going to do was make it a different kind of middleware. I will be certainly be using your code and any help with tests etc would be much appreciated
I also credit you for the work you have done with pgschemas. The way you have done your static tenant I really like.

lorinkoz:
Sounds like a good plan.

I created a gist to keep track of the ways I've found to construct the dynamic module with tenant prefixed urls.

https://gist.github.com/lorinkoz/78a908262c3a879f1d3724fe2dc64c1d

The method called `get_from_lazy_module` could be promising, if you can experiment with that yourself, maybe the solution is this way, since this method doesn't execute the original urlconf module at the time of creation.

lorinkoz:
@tomturner To clarify, you're going to use parts of this code in a PR of yours that will come at some point, and I will help you with tests at that point. Is that correct?

tomturner:
@lorinkoz Yes correct that would be great. I will be working on it soon.  I want this as an optional setting. So what I was thinking is a to add another type of middleware that way it doesn't cause existing users a problem. It looks like you been having fun with this change

lorinkoz:
Will come back in a new shape, stay tuned.

PR #371, opened 2020-04-18
add --noinput argument support to create_tenant command

lejenome:
This patch adds support for `--noinput` (or `--no-input`) argument to `create_tenant` command.
If used, it will not ask for user input even for missing fields. And in case one of these missing fields is required, the command will exit with a "Missing required field" error.

This patch will allow using the command on scripts where there is no access to terminal input or when script automation is needed.


tomturner:
@lejenome Thanks for that

PR #379, opened 2020-04-20
Fix clone_tenant command

adnathanail:
- Make everything use `field.attname` instead of `field.name` which fixes supplying foreign keys as an argument
- Make getting domain arguments work by adding "domain_"
- Print out integrity errors when they occur - I'm not 100% sure why they're being caught but it didn't think it would case a problem just printing them, and it is useful for debugging individual database design choices (e.g. I had a OneToOneField on my Client which meant I couldn't copy client fields because that would mean 2 Clients pointing to the same object, but the integrity error for that was hidden)


tomturner:
@adnathanail Thanks for that.

PR #380, opened 2020-04-20
Swap pg_authid for pg_roles

adnathanail:
- As far as I know, these tables are equivalent apart from the password column, however managed database services (Digital Ocean, and RDS) don't allow access to pg_authid


adnathanail:
Accidentally deleted the repository!

PR #391, opened 2020-04-24
Add Postgres test matrix in CI. Closes #383

atodorov:
- uses official Postgres docker images, see
  https://hub.docker.com/_/postgres
- uses GitHub actions which seems rather fast!
- uses a single version of Python b/c the point here is
  not to test the Python matrix
- uses latest Django b/c we're not testing Django compatibility here

- latest==12 ATM but this will change in future and the matrix will
pick up the next version automatically

- 9==9.6, hence 9.5 is specified explicitly as major.minor

ATM fails for `test_clone_schema()` which isn't related to the current changes.

### DIFF
--- a/.github/workflows/postgres.yml
+++ b/.github/workflows/postgres.yml
@@ -1,0 +1,45 @@
+name: postgres
+
+on:
+  pull_request:
+  push:
+    branches: master
+
+jobs:
+  test_matrix:
+    name: ${{ matrix.postgres-version }}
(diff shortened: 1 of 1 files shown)


atodorov:
@tomturner maybe you need to go to the settings of this repository to enable external contributors to execute GitHub actions. See an example here https://github.com/MrSenko/django-tenants/pull/1

tomturner:
@atodorov thanks

PR #392, opened 2020-04-25
Document differences between domains and schema names. Closes #385

### LINKED ISSUE #385: SQL_IDENTIFIER_RE doesn't seem compatible with hostname specification
In https://github.com/tomturner/django-tenants/blob/master/django_tenants/postgresql_backend/base.py we've got `SQL_IDENTIFIER_RE = re.compile(r'^[_a-zA-Z0-9]{1,63}$')` which per Postgres docs verifies that the schema name is correct. 

However debugging an issue on my end I think you can't have underscores b/c they seem to be forbidden in hostnames, see https://stackoverflow.com/a/14622263/1431647 (I didn't go through all RFCs).

Also I don't think the upper-case characters will play nicely with hostnames as well.

I was able to create a teanant with schema_name `debug_me` but when I tried visiting `debug_me.tenants.example.com` I got a 400. Trying an arbitrary domain for non-existing tenant gives me 404 (app is configured to check for this). 

@tomturner what do you think of this? I am in favor of restricting the above mentioned regex only to lower-case characters.

atodorov:
@tomturner I have come to the realization that in kiwitcms-tenants we use the schema name and sub-domain interchangeably. This poses additional restrictions as to what constitutes a valid input data. 

The proper way to deal with this on my side is to decouple the two but that's quite a lot of changes I don't have the time to deal with right now.  Even if django-tenants provided functions to validate, even convert between schemas and domain names my app isn't ready to make use of them. 

I am leaving this as documentation only so the information is accessible to others. If I come up with something else that could be useful I will contribute it back.


tomturner:
@atodorov thanks for this

PR #415, opened 2020-06-03
Add rename_schema command

adnathanail:
I wanted this for myself so I thought I'd put it here - apologies if the docs aren't quite right I haven't done much open source stuff


tomturner:
@adnathanail Thanks

PR #438, opened 2020-06-17
Subfolder routing

### LINKED ISSUE #28: Feature-Request: Change tenant-domain from subdomain to url-parameter?

### LINKED ISSUE #183: Tenant selection based on url prefix

lorinkoz:
Should eventually close  #183 and close #28


tomturner:
@lorinkoz Do you think we can merge this in.

lorinkoz:
Yes, as long as we signal it's a beta feature.

Issue #440, opened 2020-06-18
Is it worth implementing a dummy table to store the tenant info in each schema


variable:
Maybe this has been asked or a completely stupid idea but I will give a go:

Currently we can get the tenant object from the request object. But sometimes it's impossible to access this info deep within some functions that doesn't accept the request parameters. 

My idea is to create a dummy table to store some basic info about the tenant, eg. the name of it.
So that I can in anywhere to be able to identify the tenant information without needing to access the request object.

eg.
```python
def get_tenant_name():
    DummyTable.objects.first().name
```

We can use the middleware to pre-populate the values.

Totally aware of the extra database operations, hopefully the trade off is worth doing?

tomturner:
I had to do this in my project. I created a table called TenantLink. which I have one record per tenant. I populate the one row on creation of that tenant. I don’t use it very often however once or twice I needed it when the request object isn’t available. For some reason I couldn’t put a foreign key to the tenant so I store the id. Personally I wouldn’t store the name in case later you decide to change that tenant schema name.

variable:
Yeah, name was just for example. Idea is just to identify which tenant the code is running for.

jcass77:
We do this in our project as well - just include `TenantMixin` in your model class definition.

This also allows you to use the context manager to safely switch between tenants using something like:
```python
with Tenant.objects.get(schema_name='my_tenant'):
    # ...do processing here
   pass
```

variable:
> We do this in our project as well - just include `TenantMixin` in your model class definition.
> 
> This also allows you to use the context manager to safely switch between tenants using something like:
> 
> ```python
> with DummyTable.objects.get(schema_name='my_tenant'):
>     # ...do processing here
>    pass
> ```

Hello sorry I am bit lost on using the mixin, it requires to know the schema_name but the point of having the dummy table is to find out the tenant? Maybe I misunderstood something?

Cheers!

PR #441, opened 2020-06-19
TenantMixin: Allow mixin to be used with `with ... as` statement

jcass77:
Related to #440. The docstring states that the intention is to use `TenantMixin` as a context manager:

```python
"""
Syntax sugar which helps in celery tasks, cron jobs, and other scripts

Usage:
    with Tenant.objects.get(schema_name='test') as tenant:
        # run some code in tenant test
        # run some code in previous tenant (public probably)
"""
```

...but for that to work, and for the `tenant` identifier in the `as` clause to be bound to the tenant, it needs to return a copy of `self` as well.

See Python [docs](https://docs.python.org/3.8/library/stdtypes.html#contextmanager.__enter__) for details.

### DIFF
--- a/django_tenants/models.py
+++ b/django_tenants/models.py
@@ -56,6 +56,8 @@ class TenantMixin(models.Model):
         self._previous_tenant.append(connection.tenant)
         self.activate()
 
+        return self
+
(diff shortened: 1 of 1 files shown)


tomturner:
@jcass77 Thanks

PR #455, opened 2020-07-28
fix(migrations): Fix migrate_schemas not working on Django 3.1

kozlek:
Django 3.1 adds a new option for the `migrate` management command, and the Migration handler expects to have a key named `check_applied` in the options dict.
As of today, migrations crash with Django 3.1 and the master branch of django-tenants.
Adding this new option to the parser fixes the issue.


tomturner:
Thanks @kosz85. I have made a Django 3.1 branch. when Django 3.1 comes I will merge into master. This commit says I merged it to master however I reverted that to put into the branch.

PR #458, opened 2020-08-05
Add a signal that gets sent after migrations finish for a schema

mikicz:
Hi - I'll explain my motivation for this PR first and the details bellow.

In the project I'm working on we utilize materialized views to calculate statistics to prevent querying for them over and over. We're currently migrating to using django-tenants and that presented me with a problem implementing the materialized views.  In the package that I am using for the materialized views (django-pgviews), the views get synced after migrations finish. To continue having this behaviour (it's very useful, as e.g. you'd want to have the views sync after you create a tenant, no extra coding needed if runs after each migrate) I needed a way of knowing when migrations finish for each tenant, so I can recreate/refresh materialized views for the tenant, they often rely on the migrations. Turned out to be bit difficult out of the box. This is where this signal comes in.

There is current a signal called `post_schema_sync`, which however only gets called when a tenant is created (and only if it's created from calling `save()`). So to achieve the scenario I described above I had to:
  - override migrate_schemas to be able to
  - override the executor to be able to 
  - send a signal 
  - register a receiver for the signal which runs the code syncing materialized views (that bit is completely ok, just the schema needs to be active)

But with the introduction of the signal I implemented in this PR, the process would shrink down to just registering the receiver. I'm sure this is not a singular usecase, I think it could be useful for other stuff as well.

The signal gets called in the process where the migration is running, so that's why there's the conditions in the tests for executor, if the executor is multiprocessing then the signal gets sent in a different process (but it can still be caught there).

Btw, if this gets rejected - would you welcome a PR implementing customisable executors (so in my code I would be able to just override the executor which would send the signal)?

### DIFF
--- a/django_tenants/migration_executors/base.py
+++ b/django_tenants/migration_executors/base.py
@@ -3,6 +3,8 @@ import sys
 from django.db import transaction
 
 from django.core.management.commands.migrate import Command as MigrateCommand
+
+from django_tenants.signals import schema_migrated
 from django_tenants.utils import get_public_schema_name, get_tenant_database_alias
 
 
@@ -47,6 +49,7 @@ def run_migrations(args, options, executor_codename, schema_name, allow_atomic=T
         pass
 
     connection.set_schema_to_public()
+    schema_migrated.send(run_migrations, schema_name=schema_name)
 
 
 class MigrationExecutor:
--- a/django_tenants/signals.py
+++ b/django_tenants/signals.py
@@ -12,6 +12,11 @@ schema_needs_to_be_sync.__doc__ = """
 Schema needs to be synced
 """
 
+schema_migrated = Signal(providing_args=['schema_name'])
+schema_migrated.__doc__ = """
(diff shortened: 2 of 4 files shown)


mikicz:
To be honest I have no idea why the tests aren't running ok? It seems the tests don't even get launched, like it's an issue with launching the runner or something. The tests ran ok locally.

tomturner:
@mikicz I think I understand what you are trying to do. I don't have a problem with this PR as long as we can get the tests to run!   I personally want to put a single in after every migration is applied to so I can update a process bar as a tenant is being created. Ideally I would love to know how many migration need to applied so I can do a percentage.

mikicz:
I rebased on the latest master, let's see if that does it.

I could include the idx and count parameters of run_migrations in the signal call, so at least for the standard executor you would have the information for the progress bar?

tomturner:
@mikicz Thanks.  I will be manually testing / merging within a day or so. I will also be adding the stuff I need for my progress bar.

PR #462, opened 2020-08-14
Update docs - BaseTenantCommand

io-ma:
It was not clear where to import BaseTenantCommand from, I think an example is helpful


tomturner:
@io-ma Thanks

Issue #508, opened 2020-12-17
Tenants and transactions.


BITSOLVER:
I've started looking at django-tenants and really like it.

I'm just trying to figure out how transactions behave within schemas.

If I create a transaction on tenant_1 then would this block a transaction on tenant 2, until transaction 1 completes?

I am thinking on placing django-celery-results and django-celery-beat database scheduler in the public schema.  If a transaction in tenant_1 rolled back then would this also roll back data in the public schema?  If it did then I think I would need to place celery related apps in a separate database?

BITSOLVER:
I figured, after some experimenting, the best way is to put the django clelery apps into a separate database and create a separate database router.

I also had to modify to django_tenates database router to exclude django celery apps.  I was going to replace the django_tenants database router with by own, but the django_tenats app config checks for it in the ready() method.  Could we add a setting to specify an alternative router, but default it?

tomturner:
Closing as part of a backlog tidy-up — not because the report was wrong, but because it can no longer be acted on as it stands.

This has been inactive for some years and a great deal has changed underneath it: `clone_schema` has been replaced wholesale from upstream pg-clone-schema, psycopg3 is now the default driver, and the supported range is Django 4.2 / 5.2 / 6.0 on PostgreSQL 14–17. There's a fair chance this no longer reproduces.

If it still happens on **3.12.0**, please reopen with the django-tenants and Django versions, the PostgreSQL version, and the steps to reproduce — with that I can actually chase it down.

Thanks for reporting it.

PR #510, opened 2020-12-21
Added setting 'TENANT_SYNC_ROUTER' #508

BITSOLVER:
Added setting so tenant sync router can be overridden / replaced.


tomturner:
@BITSOLVER

tomturner:
@BITSOLVER thanks

PR #514, opened 2021-01-02
Update main.py

appzoid:
Remove the "return request" line from process_request of middleware class TenantMainMiddleware

with the return statement its causing:
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/wsgiref/handlers.py", line 137, in run
    self.result = application(self.environ, self.start_response)
  File "/venv/lib/python3.7/site-packages/django/contrib/staticfiles/handlers.py", line 76, in __call__
    return self.application(environ, start_response)
  File "/venv/lib/python3.7/site-packages/django/core/handlers/wsgi.py", line 133, in __call__
    response = self.get_response(request)
  File "/venv/lib/python3.7/site-packages/django/core/handlers/base.py", line 129, in get_response
    response._resource_closers.append(request.close)
AttributeError: 'WSGIRequest' object has no attribute '_resource_closers'

### DIFF
--- a/django_tenants/middleware/main.py
+++ b/django_tenants/middleware/main.py
@@ -44,7 +44,6 @@ class TenantMainMiddleware(MiddlewareMixin):
         request.tenant = tenant
         connection.set_tenant(request.tenant)
         self.setup_url_routing(request)
-#         return request
 
     def no_tenant_found(self, request, hostname):
         """ What should happen if no tenant is found.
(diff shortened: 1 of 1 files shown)


tomturner:
I would the line to be removed not just commented out

tomturner:
Thanks

PR #519, opened 2021-01-07
Move the extra extensions validations in utils and clone the django connection  after the extra extension check.

adrienrusso:
In the app.py there is a check that validates if the extra extensions are properly configured.

I have some concerns about this check:
- It prevents the use of a django command if the db is not yet created.
- It opens a connection that will not be reused and will remain inactive since the end of your connection time.

There are many solutions to solve the problem of idle connection problem:
- The first one is to close the connection after the execution of the queries to make sure that the connection will not remain idle.
- The second is to remove this check completely as it is not recommended to make a database query in the `apps.ready' file.
- Finally, we could do this check when a new connection is created using the signal created by the django database.

Let me know what you think is the best solution.

image

Idle connections when you start a server before the fix:
image

Fixed version:
image

### DIFF
--- a/django_tenants/apps.py
+++ b/django_tenants/apps.py
@@ -1,7 +1,7 @@
 from django.conf import settings
 from django.apps import AppConfig
 from django.core.exceptions import ImproperlyConfigured
-from django_tenants.utils import get_public_schema_name, get_tenant_model
+from django_tenants.utils import get_public_schema_name, validate_extra_extensions
 
 
 recommended_config = """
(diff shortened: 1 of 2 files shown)


tomturner:
Thanks I will review this

tomturner:
@adrienrusso thanks

adrienrusso:
@tomturner Do you plan to do a release soon?

tomturner:
Yes I will release soon

PR #529, opened 2021-02-08
Django 3.2

KPilnacek:
Add Django 3.2 to test matrix

### DIFF
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -13,8 +13,8 @@ jobs:
       fail-fast: true
       matrix:
         # always use latest Python b/c the point here is to test PostgreSQL compatibility
-        python-version: [3.8]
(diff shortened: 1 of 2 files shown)


tomturner:
Thank you very much. I will not push this until Dajngo 3.2 is out. Also it will need to pass the tests,

KPilnacek:
@tomturner I was working on the PR still (I don't know, if there is a better way to run the tests) so sorry for a premature PR. The tests are passing now.

Actually, what is the reason to not include Django 3.2? Many other django extensions are more benevolent in this manner:
### <4.0
* django-bootstrap3
### no upper limit 
* django-cache-lock
* django-dramatiq
* drf-nested-routers
* django-crum
* dj-rest-auth
* django-basicauth
* django-deprecate-fields
* django-cors-headers
* django-debug-toolbar
* django-filter
* djangorestframework

PR #542, opened 2021-02-27
Update admin support to handle apps with a customized label

wynnw:
- This changes the templatetags for the admin template to correctly
identify apps in settings.SHARED_APPS if they have customized their
label and don't follow the default pattern.


tomturner:
Thanks @wynnw

Issue #553, opened 2021-03-24
management commands require a running db when PG_EXTRA_SEARCH_PATHS is set.


dperetti:
Because [the check](https://github.com/django-tenants/django-tenants/blob/25eca1ce28bee5e36bfd6ac19127ab75e14eb304/django_tenants/apps.py#L38) that is done in `DjangoTenantsConfig` (still done in `master` despite refactoring), all management commands require an up and running database when PG_EXTRA_SEARCH_PATHS is set.

This breaks `collectstatic`, etc in CI...

adrienrusso:
I had the same concern as you that I addressed in this PR https://github.com/django-tenants/django-tenants/pull/519

adrienrusso:
I opened a PR for that https://github.com/django-tenants/django-tenants/pull/568. The fix was not included in my previous one.

PR #568, opened 2021-04-12
Add a switch to disable `PG_EXTRA_SEARCH_PATHS` check +  docs/install.rst cleanup

adrienrusso:
Fix the issue:
https://github.com/django-tenants/django-tenants/issues/553

If the database is not ready and you run a command, the command fails because of the extension check.

I think we should leave the choice to the user to skip or not the pg extension check as Django does for the manage commands.

### DIFF
--- a/django_tenants/utils.py
+++ b/django_tenants/utils.py
@@ -240,9 +240,10 @@ def parse_tenant_config_path(config_path):
 
 
 def validate_extra_extensions():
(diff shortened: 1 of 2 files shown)


adrienrusso:
@tomturner I think this change should be included in the next release, this is a blocker for users that run commands before the database is ready.

tomturner:
@adrienrusso thanks

PR #573, opened 2021-04-19
Fix and simplify calculation of cache_key in cached template loader

bernhardmiller:
- use calculations of parent class as base
- avoid throwing an exception if the tenant is a FakeTenant
- (brings this repo and django-tenant-schemas a bit closer together)

bonus:
- simplify instructions for testing on docker


tomturner:
@bernhardmiller

Issue #580, opened 2021-04-25
Using collectstatic_schemas in django 3.1.x fails


hanckmann:
I am using **django-tenants==3.2.1**

When using collectstatic_schemas in combination with **django 3.0.14**, the command is successful.

When using collectstatic_schemas in combination with **django 3.1.8**, the command is fails as follows:

    $ python3.8 manage.py collectstatic_schemas --schema="schemaname"
    WARNING: Using remote configuration!
    Traceback (most recent call last):
      File "manage.py", line 43, in <module>
        main(remote)
      File "manage.py", line 39, in main
        execute_from_command_line(sys.argv)
      File "/home/dev/services/python-backend/venv/lib/python3.8/site-packages/django/core/management/__init__.py", line 401, in execute_from_command_line
        utility.execute()
      File "/home/dev/services/python-backend/venv/lib/python3.8/site-packages/django/core/management/__init__.py", line 395, in execute
        self.fetch_command(subcommand).run_from_argv(self.argv)
      File "/home/dev/services/python-backend/venv/lib/python3.8/site-packages/django/core/management/base.py", line 322, in run_from_argv
        parser = self.create_parser(argv[0], argv[1])
      File "/home/dev/services/python-backend/venv/lib/python3.8/site-packages/django/core/management/base.py", line 296, in create_parser
        self.add_arguments(parser)
      File "/home/dev/services/python-backend/venv/lib/python3.8/site-packages/django_tenants/management/commands/__init__.py", line 122, in add_arguments
        self.command_instance.add_arguments(parser)
      File "/home/dev/services/python-backend/venv/lib/python3.8/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 39, in add_arguments
        parser.add_argument(
      File "/usr/local/python3.8.8/lib/python3.8/argparse.py", line 1386, in add_argument
        return self._add_action(action)
      File "/usr/local/python3.8.8/lib/python3.8/argparse.py", line 1749, in _add_action
        self._optionals._add_action(action)
      File "/usr/local/python3.8.8/lib/python3.8/argparse.py", line 1590, in _add_action
        action = super(_ArgumentGroup, self)._add_action(action)
      File "/usr/local/python3.8.8/lib/python3.8/argparse.py", line 1400, in _add_action
        self._check_conflict(action)
      File "/usr/local/python3.8.8/lib/python3.8/argparse.py", line 1539, in _check_conflict
        conflict_handler(action, confl_optionals)
[...]

PR #602, opened 2021-05-17
Fix for Using collectstatic_schemas in django 3.1.x fails

hanckmann:
This is a fix for the issue when using collectstatic_schemas in combination with django 3.1 or newer, that the command fails with: "argparse.ArgumentError: argument --skip-checks: conflicting option string: --skip-checks".

See issue: #580 

This patch is copied from: https://github.com/bernardopires/django-tenant-schemas/pull/650/commits
and especially: https://github.com/bernardopires/django-tenant-schemas/pull/650/commits/c25b27666afa9958399ba8d3b428024ae41b0c02#diff-f3d37263ffe8147dde1bee2da836b530bb3cac9ea6a4f4daa2f7ab7a506e2bfa

### DIFF
--- a/django_tenants/management/commands/collectstatic_schemas.py
+++ b/django_tenants/management/commands/collectstatic_schemas.py
@@ -3,4 +3,5 @@ from django.contrib.staticfiles.management.commands import collectstatic
 
 
 class Command(TenantWrappedCommand):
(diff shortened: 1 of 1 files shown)


tomturner:
@hanckmann thanks

PR #668, opened 2021-10-08
Use quotes around database username

atodorov:
to prevent errors like this one:

```
File "/venv/lib64/python3.8/site-packages/django/db/backends/utils.py", line 82, in _execute
    return self.cursor.execute(sql)
psycopg2.errors.SyntaxError: syntax error at or near "-"
LINE 695: ...lone_schema(text, text, boolean, boolean) OWNER TO kiwi-usr;
```

Notes: `kiwi-usr` is the username for one of my production databases and clone_tenant fails there. I think a simple


atodorov:
I will look into the current failures to see if I can figure out something.

atodorov:
> Thanks you for this. The problem with the CI was first due to the matrix. The fact that the old Django version didn't work with 3.9. At the time I didn't have time to fix it. It's now fixed as I have removed unsupported version of Django from the testing.

Thanks. Would you consider bumping the version on PyPI ? I can do that as well if you don't have the time, but I'll need access.

PR #674, opened 2021-10-22
Improvement: Explicitly store the current `search paths' in the connection.

adrienrusso:
The goal of this PR is to improve the way the current 'search_path' is handled in the database wrapper when `TENANT_LIMIT_SET_CALLS` is True

Instead of having a boolean, I replaced `search_path_set` with `search_path_set_schemas` which is the actual `search_paths` of the connection. 

This will help third-party libraries implement additional checks to avoid potential leaks.

I also extracted the generation of the `search_paths` into a private method to improve readability and flexibility for third-party libraries.

[This is a suggestion for improvement.]

Thanks.


tomturner:
@adrienrusso Thankss

PR #675, opened 2021-10-22
Docs: Remove django.contrib.contenttypes from TENANT_APPS. Closes #273

### LINKED ISSUE #273: Missing permission when tenant user tries to create object from TENANT_APPS
In short I have a public schema which holds django.contrib.auth together will all other apps. Basically SHARE_APPS == INSTALLED_APPS. Then TENANT_APPS is a subset of that:

```
        TENANT_APPS = [
            # this is from the documentation

            # the rest is a subset of INSTALLED_APPS, e.g. models which tenants should have access to, aka seeing only their own data
            'django.contrib.sites',

            'attachments',
            'django_comments',
            'modernrpc',
            'simple_history',

            'tcms.core.contrib.comments.apps.AppConfig',
            'tcms.core.contrib.linkreference',
            'tcms.management',
            'tcms.testcases.apps.AppConfig',
            'tcms.testplans.apps.AppConfig',
            'tcms.testruns.apps.AppConfig',
        ]
```

With this setup a regular user can access the tenant and view objects if they exist. However they are not able to create such objects. For example this user is missing the `testcases.add_testcase` permission.

I have narrowed down this to:

```
SELECT 
    "auth_permission"."id",
    "auth_permission"."name", 
    "auth_permission"."content_type_id", 
    "auth_permission"."codename" 
FROM "auth_permission" 
INNER JOIN "auth_group_permissions" ON ("auth_permission"."id" = "auth_group_permissions"."permission_id") 
INNER JOIN "auth_group" ON ("auth_group_permissions"."group_id" = "auth_group"."id") 
INNER JOIN "auth_user_groups" ON ("auth_group"."id" = "auth_user_groups"."group_id") 
INNER JOIN "django_content_type" ON ("auth_permission"."content_type_id" = "django_content_type"."id") 
WHERE "auth_user_groups"."user_id" = 3428 
ORDER BY "django_content_type"."app_label" ASC, "django_content_type"."model" ASC, "auth_permission"."codename" ASC
```

That is the SQL query which Django issues internally to figure out what kinds of permissions does the current user have assigned to them.

When issued against public schema it returns 235 objects, when issued against a tenant schema it returns 164 objects. The ones I'm looking for (with `@permissions_required`) are missing.

From what I can tell the `auth_permission` table has a foreign key to `django_content_type` and the contents of public.django_content_type and tenant.django_content_type are of course different. In this scenario the INNER JOIN above filters out permissions for which the content type ID doesn't match.

In the source code I was able to find comments about clearing ContentType cache and similar explanations about different IDs for the same model on the 2 schemas.

Given that I want to have all users and permissions controlled from the public schema is there any compelling reason to have 'django.contrib.contenttypes' in TENANT_APPS ?

atodorov:
I've been running without this setting in TENANT_APPS for 2.5 years now and haven't seen any issues so I propose removing it. 

Maybe it is only necessary if you have other TENANT_APPS which require it, IDK. Although in https://github.com/django-tenants/django-tenants/issues/273#issuecomment-489220660 I also point out that django_comments worked as expected so not sure. 

Proposing doc update b/c other comments on 273 seem to agree that we don't need contenttypes in TENANT_APPS.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -118,9 +118,6 @@ To make use of shared and tenant-specific applications, there are two settings c
     )
 
     TENANT_APPS = (
-        # The following Django contrib apps must be in TENANT_APPS
-        'django.contrib.contenttypes',
-
         # your tenant-specific apps
         'myapp.hotels',
         'myapp.houses',


tomturner:
@adrienrusso thanks

PR #676, opened 2021-10-28
Refactor database backend for easier subclassing

kozlek:
Use introspection_class hook to override DatabaseWrapper introspection instead of directly calling our own `DatabaseSchemaIntrospection` to enable easier backend subclassing.


tomturner:
@kozlek Thanks

tomturner:
Sorry. I had to revert this change as it cause my project to stop working.  I was getting an error when when i was linking from a tenant table to public tenant. Without this change my system works fine.

PR #677, opened 2021-10-28
Enable to order tenants when migrating schemas

mikicz:
In our usage of this package at my company we have "active" and "deactivated" tenants and when migrating we'd like to migrate active tenants first - without overriding the entire command it's not currently possible, so adding a way to do that.


tomturner:
Thanks I will merge at some point

tomturner:
@mikicz Thanks

PR #678, opened 2021-10-28
Fix executor conditions in tests

mikicz:
`get_executor()` returns a class, so the conditions currently don't work in the tests

### DIFF
--- a/django_tenants/tests/test_tenants.py
+++ b/django_tenants/tests/test_tenants.py
@@ -699,11 +699,21 @@ class SchemaMigratedSignalTest(BaseTestCase):
         """
         executor = get_executor()
 
+        print("**** DEBUG: existing tenants=", get_tenant_model().objects.all())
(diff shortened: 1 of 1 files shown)


tomturner:
I will not merge if the tests don't work

mikicz:
Yeah fair - just wanted to bring attention to this, that the tests are not working properly in those cases which I noticed in the other branch I submitted. If I find time I might to fix the tests themselves

atodorov:
FTR @tomturner https://github.com/django-tenants/django-tenants/actions/runs/3243016823/jobs/5317126799 from #852 (commit also included in this PR) shows that the original tests are bogus like @mikicz has suggested.  What happens is none of the conditions around executor succeeds and b/c there are no other assertions the test function reports PASS. That's why we didn't know anything was off.

I'll see if I can figure out why the rest of them are failing now that the conditions have been fixed.

PR #690, opened 2021-11-12
Custom Tenant Apps Feature

sunegowtham:
A feature to allow dynamically setting apps applicable for each tenant, an extension of Multi-type tenants. Set up the following in `settings.py`:
1. Set `HAS_CUSTOM_TENANT_APPS` to `True`.
2. Create a list `SHARED_APPS` as usual.
3. Instead of `TENANT_APPS` make two lists `MANDATORY_TENANT_APPS` and `OPTIONAL_TENANT APPS`. 
4. `MANDATORY_TENANT_APPS` consists of all apps required for all tenants such as `django.contrib.admin`.
5. `OPTIONAL_TENANT_APPS` should be a list of dictionaries consisting of two properties: `app` - the app name, and `dependencies` - a list of apps that this app depends on. (Then the admin user won't be able to add this app to a tenant without including its dependencies) For example,
`OPTIONAL_TENANT_APPS = [  
    {'app': 'authors', },
    {'app': 'books', 'dependencies': ['authors'], }
]`
6. Add the following code:
`TENANT_APPS = MANDATORY_TENANT_APPS + [app['app'] for app in OPTIONAL_TENANT_APPS]`
`INSTALLED_APPS = list(SHARED_APPS) + [app for app in TENANT_APPS if app not in SHARED_APPS]`

A new Tenant Mixin `TenantWithCustomAppsMixin` manages migrations when (such as remove migrations and tables when an app is removed from apps of the tenant, and migrate tables when a new app is added) `auto_migrations_on_apps_change` property is set to true.

One possible improvement — the URLs can be made resolved on a tenant-specific basis. One way is perhaps to include a URLs property on `OPTIONAL_TENANT_APPS ` which can be used by a middleware to only allow relevant URLs for the tenant.

A sample application is [here](https://github.com/django-tenants/django-tenants/files/7527483/Python.zip).

### DIFF
--- a/django_tenants/apps.py
+++ b/django_tenants/apps.py
@@ -3,7 +3,6 @@ from django.apps import AppConfig
 from django.core.exceptions import ImproperlyConfigured
 from django_tenants.utils import get_public_schema_name, validate_extra_extensions
 
-
 recommended_config = """
 Warning: The recommend way of setting django tenants up is shown in the documentation.
 Please see https://django-tenants.readthedocs.io/en/latest/install.html?highlight=#configure-tenant-and-shared-applications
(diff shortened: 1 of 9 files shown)


tomturner:
@sunegowtham  I like the idea I will review at some point.

sunegowtham:
@tomturner just wanted to check if you got the change to review it

tomturner:
@amirthya Would you be able to check some tests for this? If you do I will merge it.

amirthya:
Hi @viktorisacenko, I haven't been able to do the tests yet. @tomturner could you suggest what tests are needed? I can work on them.

H0nGzA1:
My requirement is to need this function, thank you, hope to merge

PR #698, opened 2021-12-03
skipping checks in collect_static (required for django 3.2)

hanckmann:
This is a fix for the collect_static issue when using django 3.2.

It is using the code from here: https://github.com/django-tenants/django-tenants/issues/580#issuecomment-935974452

I have tried to run the tests, but the description for how to run them (using docker) where not complete or working. So I am not sure if all tests pass. However, I have tested it empirically and the fix works.


tomturner:
thanks to @m7salam and @hanckmann

PR #704, opened 2021-12-20
Allow schema names with special chars

### LINKED ISSUE #688: Schema name
When I create a schema, I have noticed that it does not accept non `[a-z][0-9_`  characters, but when you look at the [PostgreSQL Schema Create docs](https://www.postgresql.org/docs/13/sql-createschema.html) there is no reason why it shouldn't. In fact I created schema manually with the name `my-test-schema_933333` and I had no problem. 

Are you able to look into this? Thanks

Dresdn:
Closes #688


tomturner:
@Dresdn thanks

Issue #702, opened 2021-12-12
Error while migrating schema after creating tenant


AymenRc:
Have been trying to find the error but I was not able to do it. After creating a tenant using django-admin, and using the `auto_create_schema=True` it should normally create the schema and migrate it, but all it does is just create the schema with the table django-migrations. The log I got is as follows:

`[12/Dec/2021 20:06:17] ?[m"GET /admin/clients/client/add/ HTTP/1.1" 200 25405?[0m
[12/Dec/2021 20:06:18] ?[m"GET /admin/jsi18n/ HTTP/1.1" 200 3195?[0m
[12/Dec/2021 20:06:18] ?[36m"GET /static/jazzmin/js/change_form.js HTTP/1.1" 304 0?[0m
[1/1 (100%) ?[31mstandard?[0m:?[31mtest2?[0m] ?[31m=== Starting migration?[0m
[1/1 (100%) ?[31mstandard?[0m:?[31mtest2?[0m] ?[36;1mOperations to perform:?[0m
[1/1 (100%) ?[31mstandard?[0m:?[31mtest2?[0m] ?[1m  Apply all migrations: ?[0madmin, auth, clients, contenttypes, sessions
[1/1 (100%) ?[31mstandard?[0m:?[31mtest2?[0m] ?[36;1mRunning migrations:?[0m
[1/1 (100%) ?[31mstandard?[0m:?[31mtest2?[0m]   No migrations to apply.`

Thank you in advance for your help!

T4cceptor:
Same issue here, I assume this is related to the latest version 3.4.0 - as a colleague of mine tested it with version 3.3.4 and it worked fine, but I couldn't figure why it would happen (tbh I'm not that deep into the code...)

JupiterThreads:
I can confirm I'm experiencing the same issue. I can't create a tenant with django-tenant-users. The migrations aren't happening and the schema isn't switching to the new tenant schema.

paulonteri:
Same issue here.
Downgraded to version `3.3.4` to get it working again.

mikicz:
I had this issue as well with 3.4.0, the code I submitted in PR #705 fixed it for me locally

PR #705, opened 2021-12-21
Ensure django_migrations table is created in a schema before migrations begin

mikicz:
This should fix https://github.com/django-tenants/django-tenants/issues/702

I've done some digging and I think the cause of the bug in #702 is that when the migrations run for the first time, the search path is set to the new schema and public but there is no `django_tenants` . So when the migrations run, there's a query to see which migrations haven't been applied, since there is no `django_tenants` table in the new schema, it looks at the table in the public schema where all migrations have been applied, so no migrations are run. Then the `django_tenants` table is somehow actually created in the new schema. This means that next time the migrations run, it correctly picks up the empty table and runs all migrations.

This PR makes sure that the `django_tenants` exists before running any migrations, fixing the issue.

### DIFF
--- a/django_tenants/migration_executors/base.py
+++ b/django_tenants/migration_executors/base.py
@@ -3,6 +3,7 @@ import sys
 from django.db import transaction
 
 from django.core.management.commands.migrate import Command as MigrateCommand
+from django.db.migrations.recorder import MigrationRecorder
 
 from django_tenants.signals import schema_migrated, schema_migrate_message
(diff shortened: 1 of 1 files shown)


mikicz:
Thanks for merging!

Only now see that I used `django_tenants` 4 times instead of `django_migrations` :smile:

tomturner:
@mikicz Thanks for that.  I though it was going to fix a bug I have with the latest release however it doesn't

Issue #647, opened 2021-08-10
if a project following this "project.app.apps.AppConfig" structure of declaring installed apps, the library doesn't work in admin template tags


m7salam:
I usually declare my apps with the full path as the following pattern
project.app.apps.AppConfig

using cookiecutter-django for example also follows the same structure.

the django_tenants doesn't resolve the correct app_label when it comes to this structure 

which makes the admin panel acts not as expected

the suggested would be to handle the apps that is registered following this structure 

project.app.apps.AppConfig

in django_tenants/templatetags/tenant.py

also using the module.__name__ to check doesn't resolve the correct way in case u are using for example allauth library or cookie-cutter for that matter as explained above

itzomen:
I don't know what you are getting wrong but it's works for me when I use this structure
project.apps.app.appConfig

Make sure you're defining your app appropriately in the settings

Issue #577, opened 2021-04-23
Not available for this type of schema


hanckmann:
Since my last update of all my packages, the admin site does not work properly anymore.
Almost all Apps in the admin site are grey and show the text "Not available for this type of schema".

The packages (and their versions) which I use are listed below. I highlighted those which might be most relevant.

argon2-cffi==20.1.0
asgiref==3.3.4
boto3==1.17.56
botocore==1.20.56
cffi==1.14.5
concurrent-log-handler==0.9.19
**Django==3.2**
django-clearcache==1.2.1
django-environ==0.4.5
django-hashids==0.4.0
django-impersonate==1.7.3
django-js-asset==1.2.2
django-mptt==0.12.0
django-redis==4.12.1
django-ses==2.0.0
**django-tenants==3.3.0**
**djangorestframework==3.12.4**
djangorestframework-simplejwt==4.6.0
future==0.18.2
gunicorn==20.1.0
hashids==1.3.1
huey==2.3.2
jmespath==0.10.0
numpy==1.20.2
pandas==1.2.4
portalocker==2.3.0
psycopg2==2.8.6
pycparser==2.20
PyJWT==2.0.1
python-dateutil==2.8.1
pytz==2021.1
redis==3.5.3
s3transfer==0.4.2
six==1.15.0
sqlparse==0.4.1
urllib3==1.26.4

Any idea what this  might be?
Do you need more information?

I use a make file to enter commands, so I am very sure that I only run the tenant aware versions of the management commands.

RohanGhodke:
Did you find out the answer?

tomturner:
Are you logging in to the admin from the shared / public url or from a tenant url?

RohanGhodke:
The issue is for both, the shared as well as tenant url, I tried including the tenant apps in shared app also still no effect. Also the migrations are applied just shows not available in django admin

tomturner:
This is working for me so can you send me the the installed apps config and
the shared apps config

RohanGhodke:
My list of shared and tenant apps

SHARED_APPS = (
    'django_tenants',
    'customers',
    'phonenumber_field',
    'accounts.apps.AccountsConfig',
    'api.apps.ApiConfig',
    'masters.apps.MastersConfig',
    'rest_framework',
    'rest_framework.authtoken',
    'django_extensions',

    'django.contrib.contenttypes',

    # everything below here is optional
    'django.contrib.auth',
    'django.contrib.sessions',
[...]

PR #718, opened 2022-01-20
Added support for providing AppConfig in app lists

sim1234:
Since version 3.3.0, admin is broken for apps that are specified as `path.to.apps.MyAppConfig` . This fixes it.

https://github.com/django-tenants/django-tenants/issues/577

https://github.com/django-tenants/django-tenants/issues/647


tomturner:
@sim1234 thanks

PR #742, opened 2022-03-06
[RFC] Add TenantDirectFolderMiddleware

FernandoKGA:
## Problem
We have just started a new project at the company that I work for, they've being using `django-tenants` for a while now, but they had some trouble dealing with tenant routing due to the fact [TenantSubfolderMiddleware](https://github.com/django-tenants/django-tenants/blob/688a7d2a001b8f8501dfdb59e1fc9ddc9f53818c/django_tenants/middleware/subfolder.py#L15) expects a prefix at the path string, so workarounds had to be done in order to get direct tenant paths.

In order to remove this complexity to get direct tenants at the path string (e.g. `https://example.com/tenant/some/route/for/it`), we started looking at the source code of this middleware to understand it and develop a new solution based on it.

## Solution
We've developed a simpler version of [TenantSubfolderMiddleware](https://github.com/django-tenants/django-tenants/blob/688a7d2a001b8f8501dfdb59e1fc9ddc9f53818c/django_tenants/middleware/subfolder.py#L15) called `TenantDirectFolderMiddleware`.

### How does it works?

1. It checks whether in the path string the first part is a tenant, we use the domain structure in a different way it was supposed to. Instead of using `tenant.example.com`, we simply use the name we're going to search for in the URL, like `tenant`. This should be taken in account at tenant creation.
2. If the tenant searched was not found, it will fallback to `public`.
3. If neither the tenant searched nor `public` is found, it will throw `TENANT_NOT_FOUND_EXCEPTION`. (If the public tenant wasn't found there is a big issue of configuration there...)
4. As the URLs are created dynamically, we need a way to tell the middleware which one he should pick, so you should define specific URLs to either public and tenant routes. After that you've to define them at `settings.py` like this:
```python
ROOT_URLCONF = "app.urls"
PUBLIC_SCHEMA_URLCONF = "app.urls_public"
```

## Known Limitations

1. You should be aware of the tenants domain names that are created, to avoid them colliding with public URLs (tenant domain has an equivalent route in `public`). This isn't done automatically.
2. This isn't tested against Multi-type tenants (we didn't have used this kind).

**We would like to know the contributors opinion about this PR.**

### DIFF
--- a/django_tenants/middleware/__init__.py
+++ b/django_tenants/middleware/__init__.py
@@ -2,6 +2,7 @@ import warnings
 
 from django_tenants.middleware.main import TenantMainMiddleware
 from django_tenants.middleware.subfolder import TenantSubfolderMiddleware
+from django_tenants.middleware.directfolder import TenantDirectFolderMiddleware
 
 
 class TenantMiddleware(TenantMainMiddleware):
--- a/django_tenants/middleware/directfolder.py
+++ b/django_tenants/middleware/directfolder.py
@@ -1,0 +1,67 @@
+from django.core.exceptions import ObjectDoesNotExist
+from django.db import connection
+from django.http import Http404
+from django.urls import set_urlconf, clear_url_caches
+from django_tenants.middleware import TenantMainMiddleware
+from django_tenants.urlresolvers import get_subfolder_urlconf
+from django_tenants.utils import (
+    get_public_schema_name,
+    get_tenant_model,
+    get_tenant_domain_model,
+)
+
+class TenantDirectFolderMiddleware(TenantMainMiddleware):
+    """
+    This middleware should be placed at the very top of the middleware stack.
+    Selects the proper tenant using the path direct folder. Can fail in
+    various ways which is better than corrupting or revealing data.
+    """
+
+    TENANT_NOT_FOUND_EXCEPTION = Http404
+
+    def __init__(self, get_response=None):
+        super().__init__(get_response)
+        self.get_response = get_response
+
+    def process_request(self, request):
+        # Short circuit if tenant is already set by another middleware.
+        # This allows for multiple tenant-resolving middleware chained together.
+        if hasattr(request, "tenant"):
+            return
+
+        connection.set_schema_to_public()
+
+        urlconf = None
+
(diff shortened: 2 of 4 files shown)


c0d33py:
**Info**: I have activate the `TenantDirectFolderMiddleware` in settings.py. but it's not working for me. would you like to mention that how can i access the tenant like this `mydomain.com/tenant1` instead of `tenant.mydomain.com`. i really wondering is that posible to access the tenant into single domain like  `mydomain.com/tenant1/`?

FernandoKGA:
@c0d33py Thanks for taking some time to test it. There are some steps needed to make this middleware work correctly.

1. You must insert it at `MIDDLEWARE` in `settings.py`, using the same principle as the other ones.
2. Configure two different `URLCONF`, as mentioned above. The `ROOT_URLCONF` and `PUBLIC_SCHEMA_URLCONF` routes, that will be separated files at your `app`: `urls.py` and `urls_public.py` having the desired routes.
3. The tenants should have the name that they will use to access the URL at the `domain` section. Example: At the creation of the `tenant1` tenant, I've used the same name as the `domain` name. Meaning that in it's creating, in the `domain` input I will use: `tenant1` instead of `tenant1.mydomain.com`.

This is the minimal configuration to have this middleware running.

If possible, could you make a step-by-step procedure so I can try to get the same error that you got? Maybe I'd missed something in the explanation.

FernandoKGA:
This middleware has no backward compatibility by now with domains with hostname like: `tenant1.mydomain.com`. But I think I could play with `hostname_from_request` from `TenantMainMiddleware` and something like the `reverse` from `TenantMixin` to allow some compatibility if the middleware needs to be disabled and return to `TenantMainMiddleware`, or even upgrade to this middleware.

What are your thoughts on this and the middleware it's self @tomturner ?

tomturner:
@FernandoKGA I will need some documentation for it if you want me to merge it.

FernandoKGA:
> @FernandoKGA I will need some documentation for it if you want me to merge it.

I've written some documentation for the middleware, it's located at `install.rst`. I would appreciate any feedback.

tomturner:

PR #751, opened 2022-03-31
Update app_labels util

mmcclelland1002:
Small change but I moved `get_app_label` function into utils and updated the  `app_labels` util so it can parse apps in the `AppConfig` format.


tomturner:
@mmcclelland1002 thanks

PR #760, opened 2022-04-12
Feature/read replica

mmcclelland1002:
### Summary
The current project I am working on in need of a read replica with `django-tenants` so I have been working with others to develop a relatively simple solution for read replica usage. After browsing some of the issues here on GH, I saw there was a desire for this feature as well. 

### The Problem
When `django-tenants` switches the connection, it only does so for the default database that is defined in `settings.py`. If a read replica is in place, the read replica will stay on the public schema even when the connection is switched to a tenant on the default database.
There are multiple places where the database connection is switched and would need to be overridden if we wanted to tackle this at the application or middleware level. ie; The migration executors, context managers, `connection.set_schema` - all of which could be spread throughout the codebase which would be a hassle to override.

### Proposed Solution
In django_tenants.postgresql_backend.base.py , when the new schema is set in the `self.settings_dict` in the `set_tenant` method before the connection is switched, we also do the same for the read replica database with some simple logic:
```
if READ_REPLICA_DB and DEFAULT_DB:
    if self.settings_dict["DATABASE"] == DEFAULT_DB and connections[READ_REPLICA_DB].schema_name != self.schema_name:
        connections[READ_REPLICA_DB].set_schema(self.schema_name)
```
This will switch the read replica and keep it in sync whenever the default / write database connection has been switched.
Also, since this is done at the DatabaseWrapper level, the migration executors, context managers, and `connection.set_schema` all work with read replicas with this change.

I have been testing this out and it seems to be working without a hitch so far. This solution is just a start but I hope it can be merged in so more can contribute and help develop this into a more robust solution.

PS: I have also updated the documentation to give examples on how to setup the database router and database configurations in settings.py

Thanks!

### DIFF
--- a/django_tenants/postgresql_backend/base.py
+++ b/django_tenants/postgresql_backend/base.py
@@ -7,6 +7,7 @@ from django_tenants.postgresql_backend.introspection import DatabaseSchemaIntros
 from django_tenants.utils import get_public_schema_name, get_limit_set_calls
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ImproperlyConfigured, ValidationError
+from django.db import connections
 import django.db.utils
 import psycopg2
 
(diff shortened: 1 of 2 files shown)


tomturner:
Please see #762

mmcclelland1002:
@tomturner Awesome! If that works that will be great - much more elegant solution. Will test it out on my end. Thanks

mmcclelland1002:
@tomturner The hook seems to be working well. Will close this PR.

Also, any ETA on the release for this? Thanks in advance.

prince-tanapong:
@tomturner Do you have timeline for release this PR?

alexphelps:
@tomturner I'm also in need of this being merged in and published to pypi. Do you have a timeline for releasing this?

tomturner:
@prince-tanapong & @alexphelps I will try and release next week. I have to look at all the PR before I release

alexphelps:
@tomturner just following up on this, any idea when the next release will be published?

gr4n73:
Hello,

Just checking in on this pull request from a while back. Any updates on when this will be published?

PR #768, opened 2022-05-04
Urlresolver optimization when using sub-folders

szymonbalasz:
Just a minor optimization to reduce the number of duplicate queries when using Django Tenants with sub-folders

I was using Django Toolbar to look through the views in my DRF project. When taking a closer look at my queries I saw that something in middleware was querying the same domain multiple times per request (13-20 times on average, when just navigating simple views and serializers)  

The tenant_prefix method in urlresolvers was being called multiple times and each time it ran the objects .get() method.

This behaviour only occurs when using sub-folders. There were no dubplicate queries when using the sub-domain approach.

### DIFF
--- a/django_tenants/urlresolvers.py
+++ b/django_tenants/urlresolvers.py
@@ -30,10 +30,10 @@ class TenantPrefixPattern:
         subfolder_prefix = get_subfolder_prefix()
         try:
             # Store the domain to avoid multiple DB hits for the same domain in the same request
-            if self.cached_domain is not None \
-                    and self.cached_domain.domain == connection.tenant.domain_subfolder  \
-                    and self.cached_domain.tenant.schema_name == connection.schema_name:
-                        domain = self.cached_domain
+            if (self.cached_domain is not None and 
+                    self.cached_domain.domain == connection.tenant.domain_subfolder and 
+                    self.cached_domain.tenant.schema_name == connection.schema_name):
+                domain = self.cached_domain
 
             else:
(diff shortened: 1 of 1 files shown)


django_tenants/urlresolvers.py
```diff
@@ -22,16 +22,25 @@ def reverse(viewname, urlconf=None, args=None, kwargs=None, current_app=None):
...
     def tenant_prefix(self):
         _DomainModel = get_tenant_domain_model()
         subfolder_prefix = get_subfolder_prefix()
         try:
-            domain = _DomainModel.objects.get(
-                tenant__schema_name=connection.schema_name,
-                domain=connection.tenant.domain_subfolder,
-            )
+            # Store the domain to avoid multiple DB hits for the same domain in the same request
+            if self.cached_domain is not None \
```
cclauss, line 33: PEP8 recommends avoiding backslash line continuation in Python because whitespace to the right of the backslash breaks the script on a change that is invisible to the reader.  Use parentheses instead.  https://peps.python.org/pep-0008/#maximum-line-length

szymonbalasz: Thanks for the link! Been trysting the PyCharm auto-formatter a bit too much there.

I have refactored that bit to conform to pep8 standards now.

FernandoKGA:
@szymonbalasz Is this code safe against multiple concurrent requests from different tenants? We once had a customized middleware for `django-tenants` that had a property at class level, not request level and caused information leak between tenants.

PR #778, opened 2022-06-06
Always check if db search path was been updated

hugobranquinho:
Make search path validation more resilient:
- Always check if we have differences on selected schema
- Watch `DatabaseWrapper.connection` update to reset search path

Question:
If `TENANT_LIMIT_SET_CALLS` setting improve performance and no tradeoff is made, why this setting exists?
Don't have the full context of why this is a setting, but from what i saw in the code, looks like it could be just how the lib works?

### DIFF
--- a/django_tenants/postgresql_backend/base.py
+++ b/django_tenants/postgresql_backend/base.py
@@ -68,9 +68,11 @@ class DatabaseWrapper(original_backend.DatabaseWrapper):
 
         self.set_schema_to_public()
 
-    def close(self):
-        self.search_path_set_schemas = None
-        super().close()
+    def __setattr__(self, name: str, value) -> None:
+        if name == "connection":
+            # If connection is updated, reset search path
+            super().__setattr__("search_path_set_schemas", None)
+        return super().__setattr__(name, value)
 
     def set_tenant(self, tenant, include_public=True):
         """
@@ -134,18 +136,18 @@ class DatabaseWrapper(original_backend.DatabaseWrapper):
         else:
             cursor = super()._cursor()
 
-        # optionally limit the number of executions - under load, the execution
-        # of `set search_path` can be quite time consuming
-        if (not get_limit_set_calls()) or not self.search_path_set_schemas:
-            # Actual search_path modification for the cursor. Database will
-            # search schemata from left to right when looking for the object
-            # (table, index, sequence, etc.).
-            if not self.schema_name:
-                raise ImproperlyConfigured("Database schema not set. Did you forget "
-                                           "to call set_schema() or set_tenant()?")
+        # Actual search_path modification for the cursor. Database will
+        # search schemata from left to right when looking for the object
+        # (table, index, sequence, etc.).
+        if not self.schema_name:
+            raise ImproperlyConfigured("Database schema not set. Did you forget "
+                                       "to call set_schema() or set_tenant()?")
 
-            search_paths = self._get_cursor_search_paths()
+        search_paths = self._get_cursor_search_paths()
 
+        # limit the number of executions of `set search_path` if no changes on db wrapper
+        # this can be quite time consuming
+        if self.search_path_set_schemas != search_paths or not get_limit_set_calls():
             if name:
                 # Named cursor can only be used once
                 cursor_for_search_path = self.connection.cursor()


adnathanail:
I was also wondering this!

mikicz:
When I turned on `TENANT_LIMIT_SET_CALLS` in a project at my work, it definitely broke a couple of things that I needed to fix before rolling it out, I'm not sure exactly why it broke things but it did, just needed to change ways of activating tenants and that mostly fixed it I think?

So IMO removing this functionality this quickly without deprecation is not the best idea - it might break things silently for people with a strong testsuite.

adnathanail:
That’s good to know! What sort of things did it affect?

hugobranquinho:
@mikicz do you have a db connection poll?

mikicz:
I'm pretty sure we don't use a connection pool.

We have heavily modified tests, using pytest with class support

I had to change this (and similar)
```
    @pytest.fixture(scope="function", autouse=True)
    def activated_schema(self, tenant):
        connection.set_tenant(tenant)
```
to
```
    @pytest.fixture(scope="function", autouse=True)
    def activated_schema(self, tenant):
       with tenant_context(tenant):
            yield
```

Then we have this utility function which sets context for a function based on the parameter of the function, had to change from

```
def _auto_context(function, parameter_name, is_id=False):
    @wraps(function)
    def wrapper(*args, **kwargs):
        if parameter_name in kwargs:
            tenant = kwargs[parameter_name]
        else:
            signature = inspect.signature(function)
            parameters = list(signature.parameters)
            if parameter_name not in parameters:
                raise ValueError(f"Incorrect usage, {parameter_name} not in {parameters}")
            tenant = args[parameters.index(parameter_name)]

        if is_id:
            tenant = Tenant.objects.only("schema_name").get(pk=tenant)

        if connection.schema_name == tenant.schema_name:
            result = function(*args, **kwargs)
        else:
            with tenant:
                result = function(*args, **kwargs)

        return result

    return wrapper


# a decorator for functions (and methods) which activates an tenant schema
# based on the parameters of the function (if called from tests e.g.)
auto_tenant_context = partial(_auto_context, parameter_name="tenant", is_id=False)
auto_tenant_id_context = partial(_auto_context, parameter_name="tenant_id", is_id=True)
```

To this
```

def _auto_context(function, parameter_name, is_id=False):
    @wraps(function)
    def wrapper(*args, **kwargs):
        if parameter_name in kwargs:
            tenant = kwargs[parameter_name]
        else:
            signature = inspect.signature(function)
            parameters = list(signature.parameters)
            if parameter_name not in parameters:
                raise ValueError(f"Incorrect usage, {parameter_name} not in {parameters}")
            tenant = args[parameters.index(parameter_name)]

        if is_id:
            tenant = Tenant.objects.only("schema_name").get(pk=tenant)

        if connection.schema_name == tenant.schema_name and (
            isinstance(connection.search_path_set_schemas, list)
            and connection.search_path_set_schemas[0] == tenant.schema_name
        ):
            result = function(*args, **kwargs)
        else:
            with tenant:
                result = function(*args, **kwargs)

        return result

    return wrapper


# a decorator for functions (and methods) which activates an tenant schema
# based on the parameters of the function (if called from tests e.g.)
auto_tenant_context = partial(_auto_context, parameter_name="tenant", is_id=False)
auto_tenant_id_context = partial(_auto_context, parameter_name="tenant_id", is_id=True)
```

It seems that in some circumstances the schema name was set even when the schema was not actually set in `connection.search_path_set_schemas`

tartieret:
We are experiencing issues due to having TENANT_LIMIT_SET_CALLS set to True. From time to time, a valid request (with a valid tenant domain) fails, due to the fact that the connection defaults to the public schema. I am still investigating the root cause. As I noticed in this PR that the logic to re-initialize a connection is modified to make sure that the search path will be reset, is this actually an issue in the current version of the package ? And if so, is there a way we could merge this fix? I am happy to help

PSU3D0:
Integrating this into a legacy Django site. There's a `SET SEARCH_PATH` on **every** database query now (as indicated by query count assertion checks). Was there a reason this decision was made?

hugobranquinho:
> Integrating this into a legacy Django site. There's a SET SEARCH_PATH on every database query now (as indicated by query count assertion checks). Was there a reason this decision was made?

@PSU3D0 What do you mean by integrating "this"? this PR or django-tenants lib?

Issue #779, opened 2022-06-09
Django 4.1 breaks migrate_schemas (KeyError: 'prune')


auirarrazaval:
In Django 4.1 a new option was added (`--prune`) which breaks the `migrate_schemas` (`django-tenants==3.4.2`). Thanks to [ this quick reply in stackoverflow](https://stackoverflow.com/a/72553621/16384621) I realize that it's probably something similar to what happened in  #259 a few years back when the `--plan` option was added to Django 2.2. For completeness, I'll also post the traceback here:

```
[standard:public] === Starting migration
Traceback (most recent call last):
  File "$directory/./manage.py", line 22, in <module>
    main()
  File "$directory/./manage.py", line 18, in main
    execute_from_command_line(sys.argv)
  File "$venv/lib/python3.9/site-packages/django/core/management/__init__.py", line 446, in execute_from_command_line
    utility.execute()
  File "$/lib/python3.9/site-packages/django/core/management/__init__.py", line 440, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "$venv/lib/python3.9/site-packages/django/core/management/base.py", line 402, in run_from_argv
    self.execute(*args, **cmd_options)
  File "$venv/lib/python3.9/site-packages/django/core/management/base.py", line 448, in execute
    output = self.handle(*args, **options)
  File "$venv/lib/python3.9/site-packages/django_tenants/management/commands/migrate_schemas.py", line 52, in handle
    executor.run_migrations(tenants=[self.PUBLIC_SCHEMA_NAME])
  File "$venv/lib/python3.9/site-packages/django_tenants/migration_executors/standard.py", line 11, in run_migrations
    run_migrations(self.args, self.options, self.codename, self.PUBLIC_SCHEMA_NAME)
  File "$venv/lib/python3.9/site-packages/django_tenants/migration_executors/base.py", line 53, in run_migrations
    MigrateCommand(stdout=stdout, stderr=stderr).execute(*args, **options)
  File "$venv/lib/python3.9/site-packages/django/core/management/base.py", line 448, in execute
    output = self.handle(*args, **options)
  File "$venv/lib/python3.9/site-packages/django/core/management/base.py", line 96, in wrapped
    res = handle_func(*args, **kwargs)
  File "$venv/lib/python3.9/site-packages/django/core/management/commands/migrate.py", line 188, in handle
    if options["prune"]:
KeyError: 'prune'
```

[...]

PR #780, opened 2022-06-09
Fix: add --prune option for django 4.1 compatibility

auirarrazaval:
See #779

### DIFF
--- a/django_tenants/management/commands/migrate_schemas.py
+++ b/django_tenants/management/commands/migrate_schemas.py
@@ -34,6 +34,8 @@ class MigrateSchemasCommand(SyncCommon):
         parser.add_argument('--plan', action='store_true',
                             help='Shows a list of the migration actions that will be performed.',
         )
+        parser.add_argument('--prune', action='store_true', dest='prune',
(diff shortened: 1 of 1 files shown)


tomturner:
I will only merge when 4.1 is out. Thanks

atodorov:
@tomturner thanks for the new release. However I'm not seeing 3.4.4 on PyPI yet. Can you please push the packages to PyPI as well ?

tomturner:
Done

cclauss:
I have `django-tenants v3.4.8` installed but unfortunately, when upgrading to Django v4.1, I am still getting:
```
>       if options["prune"]:
E       

.venv/lib/python3.9/site-packages/django/core/management/commands/migrate.py:188: KeyError
```
* Is there a way to verify `django-tenants.__version__` (or similar) at runtime?
* Are there other dependencies that might cause Django v4.1 to raise `KeyError: 'prune'`?
* Any other tips or tricks about how to avoid the KeyErrror when upgrading to Django v4.1?

% `pip list | grep tenant`
```
django-tenants            3.4.8
tenant-schemas-celery     2.0.0
```

PR #782, opened 2022-06-15
Fix TenantClient annotations

hugobranquinho:
`django.test.Client` inheritance have `RequestFactory`, so, defining `TenantRequestFactory(RequestFactory)` on `TenantClient` is causing some annotations issues.
This change creates a clean base class so we can use on both `TenantRequestFactory` and `TenantClient`


tomturner:
@hugobranquinho Thanks

PR #795, opened 2022-08-03
Allow all_tenants_command to run excluding public schema

bartonip:
From time to time we need to run commands on all tenants excluding the public tenant. An example of this might be rebuilding an Elasticsearch cache after a migration is executed. When we do this the vast majority of apps we rebuild are tenant specific and using the `all_tenants_command` fails as the public tenant does not have these tables.

I have added an extra `--no-public` flag to explicitly exclude the public schema from specific operations and cleaned up some of the jank that was in the management commands.

Appreciate your comments.

### DIFF
--- a/django_tenants/management/commands/all_tenants_command.py
+++ b/django_tenants/management/commands/all_tenants_command.py
@@ -21,30 +21,37 @@ class Command(BaseCommand):
         Changes the option_list to use the options from the wrapped command.
         """
         # load the command object.
+        print(argv)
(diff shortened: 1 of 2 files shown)


tomturner:
@bartonip Thank you.  If you want it merged quicker please update the docs on all_tenants_command in use.rst

bartonip:
@tomturner done mate, please review

tomturner:
@bartonip thanks

PR #803, opened 2022-08-08
Add Subdomain with Subfolder Middleware

mameen7:
### Problem

There are situations where a project will want to support both subdomain and subfolder tenant routing. To give an example, recently i was doing some functionality check on Slack, i realized that Slack is using both subdomain and subfolder tenant routing, This is just an example. Personally, I'm also working on a project which is a multi-tenant project, and I'm surely going to need both subdomain and subfolder tenant routing, but i realized that the existing Django-Tenant middlewares cannot be used together to solve this problem.

### Solution

I have developed a new middleware called `TenantSubdomainWithSubfolderMiddleware`, it's built based on the existing middlewares.

**How it works**

1. It first check the settings to make sure that the TENANT_SUBFOLDER_PREFIX settings is provided. It will throw ImproperlyConfigured exception if the settings is not provided.
2. It then check the requested URL path to see if the TENANT_SUBFOLDER_PREFIX is provided (i.e the request is trying to make use of subfolder tanent routing). If the SUBFOLDER_PREFIX is provided it will use the provided schema_name to get the tenant.
3. If a tenant could not be found using the provided schema_name in the URL path or the TENANT_SUBFOLDER_PREFIX is not included in the URL path, it will then find the tenant using subdomain if subdomain is provided.
4. If a tenant could not be found using the subdomain, it will route to the public tenant if SHOW_PUBLIC_IF_NO_TENANT_FOUND settings is provided, else it will raise the TENANT_NOT_FOUND_EXCEPTION exception.

### Note
For both subdomain and subfolder tenant routing to work, the tenant subdomain need to be passed to the `domain` field in the `Domain` model, and the subdomain prefix has to match with the schema_name of the tenant.

**example**
```
tenant = Client(schema_name='tenant1',
                name='Fonzy Tenant',
                paid_until='2014-12-05',
                on_trial=True)
tenant.save()
```
```
domain = Domain()
domain.domain = 'tenant1.my-domain.com'     # instead of just 'tenant1' as of subfolder middleware
domain.tenant = tenant
domain.is_primary = True
domain.save()
```


algebra7:
@tomturner can you or anyone review this?

serengetisunset:
+1 to this :)

tomturner:

PR #806, opened 2022-08-09
Remove deprecated url imports from example projects

algebra7:
`django.conf.urls.url()` has been deprecated and removed in Django 4.0+, but it is still imported in all `urls_public.py` files in the example projects, hence causing an import error.


atodorov (approved):
LGTM

tomturner:
@algebra7 Thanks

PR #810, opened 2022-08-12
Don't GRANT permissions to sequences when using IDENTITY columns.

### LINKED ISSUE #802: Weird DB failure after upgrade to django-tenants 3.4.4 & Django 4.1
I am using containerized Postgres as my database:

`centos/postgresql-12-centos7   latest    ead3e66c2b54   12 months ago   372MB`

this hasn't changed recently and was working fine in previous commits. 

After upgrade from django-tenants 3.4.3 to 3.4.4 (and other updates to the app which gets split into tenants, including upgrade to Django 4.1) I am seeing a failure in my automated tests:

```
psycopg2.errors.RaiseException: Action: PRIVS: Sequences  Diagnostics: line=SQL statement "GRANT SELECT ON subscriber.management_build_build_id_seq TO "kiwi";"
PL/pgSQL function clone_schema(text,text,boolean,boolean) line 606 at EXECUTE. 42P01. relation "subscriber.management_build_build_id_seq" does not exist
CONTEXT:  PL/pgSQL function clone_schema(text,text,boolean,boolean) line 669 at RAISE
```

Full logs: https://github.com/kiwitcms/tenants/runs/7725825769?check_suite_focus=true

Test which triggers this failure:
https://github.com/kiwitcms/tenants/blob/master/tcms_tenants/tests/test_views.py#L117

which fails here https://github.com/kiwitcms/tenants/blob/master/tcms_tenants/utils.py#L93 - executing `clone_tenant` command!

atodorov:
Fixes #802

Notes:

- see the issue for background analysis
- this patch seems like the easiest fix without refactoring the entire
  clone_schema() function!

Speculation:

I think that because IDENTITY columns use sequences internally (opposed
to creating sequences explicitly) information_schema doesn't contain
information about them anymore (resulting in seq_cnt = 0). However this
internal information is still present in pg_catalog which results in an
error when trying to grant the permissions.

Also possible that at the time the GRANT statement is executed the internal
sequences aren't created yet and they are only created only *AFTER* the entire
transaction is committed.

@tomturner this is the least invasive patch that I could figure out. Seems to work for me. I will try a bit more to reproduce with the upstream test suite but so far no luck.

Note: a new PyPI release would be awesome.

### DIFF
--- a/django_tenants/clone.py
+++ b/django_tenants/clone.py
@@ -59,6 +59,7 @@ DECLARE
   arec             RECORD;
   cnt              integer;
   cnt2             integer;
+  seq_cnt          integer;
   pos              integer;
   action           text := 'N/A';
   v_ret            text;
@@ -212,14 +213,14 @@ BEGIN
 
   -- Create sequences
   action := 'Sequences';
-  cnt := 0;
+  seq_cnt := 0;
   -- TODO: Find a way to make this sequence's owner is the correct table.
   FOR object IN
     SELECT sequence_name::text
       FROM information_schema.sequences
      WHERE sequence_schema = quote_ident(source_schema)
   LOOP
-    cnt := cnt + 1;
+    seq_cnt := seq_cnt + 1;
     IF ddl_only THEN
       RAISE INFO '%', 'CREATE SEQUENCE ' || quote_ident(dest_schema) || '.' || quote_ident(object) || ';';
     ELSE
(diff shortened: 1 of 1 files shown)


atodorov:
@tomturner more info & reproducers in https://github.com/kiwitcms/tenants/pull/181#issuecomment-1231016658.

I will keep debugging this until I have all tests passing and no unexplained changes to the test suite are related. Can you please review?

tomturner:
@atodorov Thanks

atodorov:
@tomturner I am still trying to reproduce the original issue on my end (which I was able to) and trying to verify the fix (see 4th commit at https://github.com/kiwitcms/tenants/pull/181) which I can't. 

For some reason I am getting a failure (it's from my app code) which I wasn't getting before. I don't think it's related, more likely some weird complexity in my testing code.

Do you mind publishing a new PyPI version though ?

PR #812, opened 2022-08-17
delaying import of ContentType until django is fully loaded

landonsilla:
I recently upgraded to `Django 3.2.14` and this is now causing a problem.
Screen Shot 2022-08-12 at 6 34 03 PM
It seems that it's trying to load the ContentType model before it's fully loaded.  So I'm changing the code here such that the model is loaded as late as possible, after django is fully initialized.

Is this a good idea? Is there a better way?


analogrithems:
I'm having the same issue with upgrading to Django 3.2.15 the ContentType loads to early

PR #814, opened 2022-08-19
Feature/get_executor hook

mmcclelland1002:
**Problem**: I was writing my own `MigrationExecutor` class to handle soft deleted tenants but ran into an issue with injecting the executor into the `migrate_schemas` command. I could've imported the `migrate_schemas` management command into my project but it seemed like overkill.

**Solution**: Added a simple hook to override this with my own custom `get_executor` function.


tomturner:
@mmcclelland1002 Thanks

PR #819, opened 2022-08-24
Add decorator to make migrations.RunPython data migrations tenant-aware

mmcclelland1002:
**Problem**: In my project, I often create data migrations using `migrations.RunPython` (https://docs.djangoproject.com/en/4.1/ref/migration-operations/#runpython). When creating a data migration that I want to run only in the public schema or only in the tenant schema, I often had to create redundant `if else` logic checking the connection schema to prevent my data migration from running on the wrong schema.

**Solution**: Create an easy, lightweight decorator to allow control on when the function executes in a migration.

Added some extra documentation to show how to use this util.


tomturner:
@mmcclelland1002 Looks good I will give it a go

mmcclelland1002:
thanks @tomturner . i've been using it in my project regularly the past few weeks and has been working well

tomturner:
@mmcclelland1002 thanks

Issue #482, opened 2020-09-30
ContentType cache should only be cleared if installed as TENANT_APPS


thenewguy:
[The backend clears the ContentType cache every time a tenant is set](https://github.com/django-tenants/django-tenants/blob/master/django_tenants/postgresql_backend/base.py#L84)

Per https://github.com/django-tenants/django-tenants/issues/171#issuecomment-399713498 it should only be cleared if ContentTypes is installed as a TenantApp

adnathanail:
This is a good point!

I've created a PR #826

PR #826, opened 2022-09-03
Only clear contenttypes cache when necessary

adnathanail:
I think this change should address issue #482

### DIFF
--- a/django_tenants/postgresql_backend/base.py
+++ b/django_tenants/postgresql_backend/base.py
@@ -86,15 +86,19 @@ class DatabaseWrapper(original_backend.DatabaseWrapper):
             EXTRA_SET_TENANT_METHOD(self, tenant)
 
         self.search_path_set_schemas = None
-
-        # Content type can no longer be cached as public and tenant schemas
-        # have different models. If someone wants to change this, the cache
-        # needs to be separated between public and shared schemas. If this
-        # cache isn't cleared, this can cause permission problems. For example,
-        # on public, a particular model has id 14, but on the tenants it has
-        # the id 15. if 14 is cached instead of 15, the permissions for the
-        # wrong model will be fetched.
-        ContentType.objects.clear_cache()
+        
+        if "django.contrib.contenttypes" in settings.TENANT_APPS:
+            # If the Django contenttypes app is installed in TENANT_APPS then the
+            # content type can no longer be cached as public and tenant schemas
+            # have different models.
+            # If this cache isn't cleared, this can cause permission problems.
+            # For example, on public, a particular model has id 14, but on the
+            # tenants it has the id 15. if 14 is cached instead of 15, the permissions
+            # for the wrong model will be fetched.
+            #
+            # If someone wants to change this, the cache
+            # needs to be separated between public and shared schemas.
+            ContentType.objects.clear_cache()
 
     def set_schema(self, schema_name, include_public=True, tenant_type=None):
         """
--- a/django_tenants/postgresql_backend/base.py
+++ b/django_tenants/postgresql_backend/base.py
@@ -88,7 +88,7 @@ class DatabaseWrapper(original_backend.DatabaseWrapper):
         self.search_path_set_schemas = None
         
         if has_multi_type_tenants():
-            TENANT_TYPES = getattr(settings, 'TENANT_TYPES')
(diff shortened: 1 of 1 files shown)


tomturner:
Thank you for this however this will not work if one is running with multi tenant types.
As TENANT_APPS doesn't exist.

Example config

`
HAS_MULTI_TYPE_TENANTS = True.

TENANT_TYPES = {
    "public": {  # this is the name of the public schema from get_public_schema_name
        "APPS": ['customers',
                 'django.contrib.admin',
                 'django.contrib.auth',
                 'django.contrib.sessions',
                 'django.contrib.messages',
                 'django.contrib.staticfiles', ],
        "URLCONF": "tenant_multi_types_tutorial.urls_public",
    },
    "type1": {
        "APPS": ['django.contrib.contenttypes',
                 'django.contrib.auth',
                 'django.contrib.admin',
                 'django.contrib.sessions',
                 'django.contrib.messages',
                 'tenant_type_one_only',
                 'tenant_type_both'],
        "URLCONF": "tenant_multi_types_tutorial.urls_type1",
    },
    "type2": {
        "APPS": ['django.contrib.contenttypes',
                 'django.contrib.auth',
                 'django.contrib.admin',
                 'django.contrib.sessions',
                 'django.contrib.messages',
                 'tenant_type_two_only',
                 'tenant_type_both'],
        "URLCONF": "tenant_multi_types_tutorial.urls_type2",
    }
}



`

adnathanail:
@tomturner good point! I've updated the PR to take this into account

tomturner:
@atodorov Can you take a look at this?  It seems OK to me.

django_tenants/postgresql_backend/base.py
```diff
@@ -86,14 +86,33 @@ def set_tenant(self, tenant, include_public=True):
...
-        # Content type can no longer be cached as public and tenant schemas
-        # have different models. If someone wants to change this, the cache
-        # needs to be separated between public and shared schemas. If this
-        # cache isn't cleared, this can cause permission problems. For example,
-        # on public, a particular model has id 14, but on the tenants it has
-        # the id 15. if 14 is cached instead of 15, the permissions for the
-        # wrong model will be fetched.
+        
+        if has_multi_type_tenants():
+            TENANT_TYPES = getattr(settings, 'TENANT_TYPES')
```
atodorov, line 91: Default should be an empty list/dict ?, otherwise if this setting isn't defined iterating the for loop below will crash.

atodorov: Further more getattr will raise AttributeError if default isn't provided.

django_tenants/postgresql_backend/base.py
```diff
@@ -86,14 +86,33 @@ def set_tenant(self, tenant, include_public=True):
...
+            no_content_types_in_tenant_apps = True
+            for schema in TENANT_TYPES:
+                if schema == get_public_schema_name():
+                    continue
+                if "django.contrib.contenttypes" in TENANT_TYPES[schema]["APPS"]:
+                    no_content_types_in_tenant_apps = False
+            if no_content_types_in_tenant_apps:
+                return
+        else:
+            TENANT_APPS = getattr(settings, 'TENANT_APPS')
```
atodorov, line 101: Same thing here, default should be an empty list.

atodorov (changes requested):
Minor changes requested.

@tomturner FTR I'm not well familiar with the behavior of the affected code. Changes look good to me but I can't tell if there would be any other side effects.

django_tenants/postgresql_backend/base.py
adnathanail: Ah thanks!

atodorov (approved):

atodorov:
Hmm, it looks like these changes actually affect something. In one of my projects updating to the latest django-tenants causes failures, see https://github.com/kiwitcms/tenants/actions/runs/3344599698/jobs/5539135976

```
django.db.utils.IntegrityError: insert or update on table "django_admin_log" violates foreign key constraint "django_admin_log_content_type_id_c4bce8eb_fk_django_co"
DETAIL:  Key (content_type_id)=(58) is not present in table "django_content_type".
```

I'm not sure whether I'm doing something incorrectly inside of my own app or if we're missing something here. I will open a PR to revert the current changes b/c the failure is critical for me and then will look more into what's happening.

Issue #831, opened 2022-09-16
Bug with Django 4.1: DETAIL:  Key (id)=(1) already exists.


atodorov:
How to reproduce:

0) Make sure you are using Django 4.1 or later and initialize the database (you need a fresh DB)
1) Create a schema/tenant which has some default records in any of the tables
2) Clone from the schema which contains records in order to create a new one. 
3) On the newly created tenant try adding a new record into the table which has records that were cloned from the source schema. 

You will get a failure telling you that the primary key for this record already exists! 

Background: 

In Django 4.1 and later the default is to create primary key fields as what Postgres calls IDENTITY columns. From my understanding IDENTITY columns are an encapsulation of the previous SEQUENCE columns. IDENTITY columns still use sequences behind the scenes. 

In #810 I've made some relevant changes but I don't think they are the source of the bug, rather they were incomplete. 

So with older Django versions, when using sequences the code around copying sequences in #810 takes care to inspect the last value of the sequence from the source and alter the newly copied target sequence to use that value. Hence the next time you want to add a record the sequence will be incremented and you will get a unique primary key.

With Django 4.1 and later PKs are created as identity columns and the schema cloning code doesn't find any sequences. Tables and records are still copied as expected but the `clone_schema` function never adjusts the values of these IDENTITY columns which leads to the duplicate ID error.

Workaround:
If you retry adding your records multiple times, eventually the value of the IDENTITY column will be incremented enough so that it will not conflict with existing records and the addition will succeed. 

TODO: 
- First I need to add tests to reliably reproduce this problem upstream and also in my downstream app
- Figure out how to actually fix it.

PR #834, opened 2022-09-18
Tests & patches for #831

atodorov:


atodorov:
FTR 
https://github.com/django-tenants/django-tenants/actions/runs/3077513472/jobs/4972431214 and 
https://github.com/django-tenants/django-tenants/actions/runs/3077513472/jobs/4972431786

successfully reproduce #831 which means the newly added test must be doing it's job.

atodorov:
@tomturner this is ready for review. The commit log & patch itself should be self-explanatory - in case we're using identity columns we need to reset them based on the number of records copied from the source schema. See also #831 for more background. 

I've also tested it locally within my application and it works there as well. 

A new version on PyPI when you get to it would be highly appreciated!

tomturner:
@atodorov I am happy do you want to merge this in yourself?

atodorov:
> @atodorov I am happy do you want to merge this in yourself?

I would love to but I can't. Looks like permissions issue:

image

tomturner:
@atodorov can you try again

PR #837, opened 2022-09-21
Set schema after creating migrations table

tienne-B:
This commit adds a duplicate "set_schema" after the verification for the creation of the django_migrations model. This is necessary for the migrate_schemas command to work properly when TENANT_LIMIT_SET_CALLS is True as the cursor for the transaction to create the migrations table would not set the search path back to the tenant.

Checking the existence of the migrations table does not have such an effect as Introspection is wrapped to use tenant-aware cursors. This is not possible however with Schemas; where the table is created.

### DIFF
--- a/django_tenants/migration_executors/base.py
+++ b/django_tenants/migration_executors/base.py
@@ -43,6 +43,10 @@ def run_migrations(args, options, executor_codename, schema_name, tenant_type=''
     # table in the public schema gets picked and no migrations are applied
     migration_recorder = MigrationRecorder(connection)
     migration_recorder.ensure_schema()
+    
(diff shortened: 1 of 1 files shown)


tomturner:
Thanks — and sorry this sat so long. The fix is correct, and it's already on master: `django_tenants/migration_executors/base.py` now has

```python
migration_recorder = MigrationRecorder(connection)
migration_recorder.ensure_schema()
connection.set_schema(schema_name, tenant_type=tenant_type)
```

which is exactly this change, arrived at independently. Closing as already fixed — thanks for reporting and diagnosing it.

PR #840, opened 2022-09-23
Introduce a new setting to specify a custom migrate command

browniebroke:
We want to customise a bit the `migrate` command in our project, but django-tenants doesn't offer any hooks to customise that currently.

This change introduces a new setting to customise the `migrate` command used by the `run_migrations()` function, defaulting to Django's (the way it works now).


tomturner:
@browniebroke Thanks you. You made it nice and easy for me by adding some test and docs.

PR #804, opened 2022-08-09
Update test matrix

atodorov:
I am chasing down bug #802 which happens for me locally so I figured better update the test matrix first to make sure it all works upstream.

PR #842, opened 2022-09-26
Fix Trove classifiers

browniebroke:
Add missing ones based on CI:
- Django 3.2 and 4.1 are supported by the package
- Python 3.10 is supported

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,13 +45,16 @@ setup(
         'Framework :: Django :: 2.2',
         'Framework :: Django :: 3.0',
         'Framework :: Django :: 3.1',
+        'Framework :: Django :: 3.2',
         'Framework :: Django :: 4.0',
+        'Framework :: Django :: 4.1',
         'Programming Language :: Python',
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -45,13 +45,16 @@
         'Framework :: Django :: 2.2',
         'Framework :: Django :: 3.0',
         'Framework :: Django :: 3.1',
+        'Framework :: Django :: 3.2',
         'Framework :: Django :: 4.0',
+        'Framework :: Django :: 4.1',
         'Programming Language :: Python',
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
```
browniebroke, line 53: Python 3.5 and 3.6 have both reached EOL, and aren't tested on CI. Should we remove it from the list? We probably ought to [set the `python_requires`](https://packaging.python.org/en/latest/guides/dropping-older-python-versions/#specify-the-version-ranges-for-supported-python-distributions) as well.

setup.py
```diff
@@ -45,13 +45,16 @@
...
         'Programming Language :: Python :: 3.7',
```
browniebroke, line 54: Python 3.7 isn't tested on CI, but it seems to be just to simplify the build matrix https://github.com/django-tenants/django-tenants/pull/804

tomturner:
Thanks @browniebroke

PR #859, opened 2022-10-28
Revert "Only clear contenttypes cache when necessary"

atodorov:
Reverts django-tenants/django-tenants#826. 

See comments in https://github.com/django-tenants/django-tenants/pull/826#issuecomment-1295352424.

Update: IMO we should revert this change until we have robust tests in the first place. I should have not merged this without tests! 

@tomturner what's your opinion about releasing a hot-fix package with this change ? 

@adnathanail is this change something that's critical for you or not?

### DIFF
--- a/django_tenants/postgresql_backend/base.py
+++ b/django_tenants/postgresql_backend/base.py
@@ -6,7 +6,7 @@ from importlib import import_module
 from django.utils.module_loading import import_string
 
 from django_tenants.postgresql_backend.introspection import DatabaseSchemaIntrospection
-from django_tenants.utils import get_public_schema_name, get_limit_set_calls, has_multi_type_tenants
+from django_tenants.utils import get_public_schema_name, get_limit_set_calls
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ImproperlyConfigured, ValidationError
 import django.db.utils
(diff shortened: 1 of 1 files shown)


tomturner:
I happy to release if you do the revert I will create the release

atodorov:
> I happy to release if you do the revert I will create the release

Let's give it until Monday to head from other interested parties and then make a decision.

adnathanail:
@atodorov it's not critical no, I was just investigating some content types issues of my own, found #482, and thought I might give it a go

adnathanail:
I've had a look at your repo and I think the cause might be that I set the default value for `TENANT_APPS` to `[]`

I'm not 100% sure, but I think that `django-tenants` will actually default `TENANT_APPS` to `INSTALLED_APPS`? (And the same for `SHARED_APPS`)

adnathanail:
In your repo `TENANT_APPS` is not explicitly set anywhere. I didn't realise that was a valid config!

atodorov:
> In your repo `TENANT_APPS` is not explicitly set anywhere. I didn't realise that was a valid config!

It isn't. You are just missing more context. The setting is configured in a different place because kiwitcms/tenants is a plugin that gets installed into a bigger application.

atodorov:
@tomturner I've merged the revert and will look into the failure next week. I'll open a PR once I have something working.

tomturner:
@atodorov So shall I release now with the revert?

atodorov:
> So shall I release now with the revert?

@tomturner yes please.

tomturner:
@atodorov OK

tomturner:
@atodorov Done

PR #882, opened 2023-01-19
GitHub Actions: Add Python 3.11 to the testing

cclauss:
Add Django v4.1 running on Python 3.11 to the testing.
https://docs.djangoproject.com/en/4.1/faq/install/#what-python-version-can-i-use-with-django


tomturner:
@cclauss thanks

PR #913, opened 2023-04-09
Add support for psycopg3

Jigsaw52:
Checks if the Django is using psycopg3. If the check fails with import error, the Django version does not support psycopg3 (< 4.2) and we fallback to psycopg2. If it succeeds, import the same version of psycopg as Django.


tomturner:
@Jigsaw52 Thanks

PR #941, opened 2023-06-08
fix: allow inferring app label from custom Django app config

hblanko:
Existing implementation of `django_tenants.utils.get_app_label()` assumes that the Django app label corresponds to the name of the main module of the app. This is not the case when `AppConfig.label` is set to a custom value, for example:

```python
# Django settings

TENANT_APPS = [
    # ...
    "path.to.myapp",
]
```

and the Django app configuration being defined as:

```python
# path/to/myapp/apps.py

from django.apps import AppConfig

class MyAppConfig(AppConfig):
    name = "path.to.myapp"
    label = "my_app_label"
```

This is a legit app configuration that strictly follows the [Django documentation on app configuration](https://docs.djangoproject.com/en/4.2/ref/applications/#application-configuration), meant to prevent app naming collisions. Without the code shipped in this PR, `django-tenants` is unable to infer the right app label, provoking the admin site to show the message "Not available for this type of schema".

This PR covers the case of https://github.com/django-tenants/django-tenants/pull/797, but with a more complete implementation.

All tests pass.

### DIFF
--- a/django_tenants/utils.py
+++ b/django_tenants/utils.py
@@ -1,12 +1,12 @@
 import os
 from contextlib import ContextDecorator
 from functools import lru_cache, wraps
+from types import ModuleType
 
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured, ValidationError
-from django.db import connections, DEFAULT_DB_ALIAS, connection
+from django.db import DEFAULT_DB_ALIAS, connection, connections
 from django.utils.module_loading import import_string
-
 
 try:
     from django.apps import apps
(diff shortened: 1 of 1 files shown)


felixrindt:
Good catch! This just came up in our project and caused some confusion. The admin page showing "Not available for this type of schema" is the only symptom we are aware of. Something else you noticed not working with a custom app_label?

tomturner:
I sure this is the same as #797

hblanko:
Is there any way I can contribute to move forward with this PR?

tomturner:
@hblanko Thanks

Issue #971, opened 2023-09-03
hostname_from_request not working when frontend in a different host


cazonda:
I'm developing a multitenant app with DRF backend and Admin LTE(HTML+Javascript+css) frontend.
The frontend is in one host and backend is in a different host. django-tenants is determining a wrong client when the request comes from a client that is hosted in a different server than the where the backend is.

Long story short: when frontend is in server A and backend in server B, the method TenantMainMiddleware.hostname_from_request(request) returns server B when it should return sever  A

Dresdn:
[TenantMainMiddleware.hostname_from_request(request)](https://github.com/django-tenants/django-tenants/blob/master/django_tenants/middleware/main.py#L25) will always return "Server B" in your example, as `request.get_host()` returns the address the request was made TO, not FROM.

If you want to get where the request was FROM (Server A in your example), then you'll need to override and examine the `request.META` object for what matches your use case.

Going to the issue, what's the use case for needing to know what Server A is?

tomturner:
Closing as part of a backlog tidy-up — not because the report was wrong, but because it can no longer be acted on as it stands.

This has been inactive for some years and a great deal has changed underneath it: `clone_schema` has been replaced wholesale from upstream pg-clone-schema, psycopg3 is now the default driver, and the supported range is Django 4.2 / 5.2 / 6.0 on PostgreSQL 14–17. There's a fair chance this no longer reproduces.

If it still happens on **3.12.0**, please reopen with the django-tenants and Django versions, the PostgreSQL version, and the steps to reproduce — with that I can actually chase it down.

Thanks for reporting it.

PR #973, opened 2023-09-07
hostname_from_request() not working well when frontend is in a differ…

cazonda:
Fixing #971
This PR adds to django-tenants the capability of determining the tenant when the request comes from an app frontend that is hosted in a server that is not the same as backend's. 

**Crurrent Behavior**: the frontend is pure html+css+js and backend is DRF+django-tenants. So when a request comes, django-tenants redirects to the public schema because the method hostname_from_request() is only returning the destination host from the request. It should redirect to the tenant schema.

**The Fix**: consists in getting the hostname from the HTTP_ORIGIN header, only recur to request.get_host() if the first fails.

### DIFF
--- a/django_tenants/middleware/main.py
+++ b/django_tenants/middleware/main.py
@@ -22,7 +22,14 @@ class TenantMainMiddleware(MiddlewareMixin):
         """ Extracts hostname from request. Used for custom requests filtering.
             By default removes the request's port and common prefixes.
         """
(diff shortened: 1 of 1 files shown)


tomturner:
This is not required you can just use your own middleware to do this.  For my project I use custom middleware.  I am expecting people to create custom middleware.

PR #976, opened 2023-09-18
chainable subfolder middleware

mkay849:
We wanted to use the `TenantSubfolderMiddleware` in combination with the `TenantMainMiddleware`. While trying to integrate this we noticed that the middlewares have some downsides as they are implemented right now:

1. Both middlewares are not chainable
2. Subfolder middleware uses domain names to identify tenants
3. Subfolder middleware uses special files to set up urlconf per tenant

What we did to overcome the forementioned issues:

To 1.
Instead of raising a tenant not found error in the `process_request` method we check if a tenant has been set in the `process_response` method after it has passed all other middlewares, thus making both middlewares chainable. Also additional middlewares could be easily be added in the same manner.

To 2.
When first trying to activate the `TenantSubfolderMiddleware` instead of the `TenantMainMiddleware` we noticed that no tenant could be identified. As it turned out we tried to access `/clients/tenant_schema_name/rest/of/url` instead of `/clients/tenant.test.com/rest/of/url` which feels kind of wrong because a domain shouldn't be used in an url. I also searched the forum and other people that use the subfolder middleware just add a domain, which is just the schema name. When people start to use tables in a way they were not intended to, something goes wrong. That's why we identify the tenant in our approach by directly using the more intuitive schema name against the tenant model's schema name directly.

To 3.
Our last point is the most disruptive change, sorry about that :)  We spotted that the subfolder uses some magic files to determine the urlconf per tenant. Instead of going that route we removed the subfolder prefix path and the tenant subfolder from request's path_info after we have identified the tenant and everything works exactly as it does with the `TenantMainMiddleware`.

We hope that these changes can be merged. If something is missing, wrong or unclear, feel free to request changes ;)


jo-bitsch:
I would be very interested in this feature. Is there any way, how I could help out?

It looks like @mkay849 already added a bunch of tests to ensure this works as expected. Do you need additional tests? Additional documentation? I'd be very glad to help out.

dwjp:
Just wanted to check with @mkay849 if you got this working in your project?

When I tried the subfolder middleware seemed to be setting the tenant to public, but there is a matching tenant subdomain, which never gets checked given this short circuit: 

```
if hasattr(request, "tenant"):
    return
```

I can probably play around to get it working but figured someone might have solved this already. Being able to support both subfolder and domain is certainly handy, especially if you'r migrating between them.

mkay849:
@dwjp Sorry for the late reply. It used to work in my project, but I migrated away from django-tenants in the meantime.
If I remember correctly the short circuit that you are mentioning has been used like this:
- Chain the middlewares in the order of preference: Domain over subfolder or subfolder over domain
- If one middleware is able to match a tenant then it sets it
- Every chained tenant middleware checks if the tenant has been set and skips determination if it is already set

tomturner:

PR #987, opened 2023-10-26
ability to set schema for management command

granteagon:
Sometimes one might want a management command that always runs on a particular schema.  This can be done by setting `schema_name` on the child class.

```python
class MyPublicCommand(BaseTenantCommand):
    schema_name = 'myfavschema'
    # this schema will only run on public
    # can be overridden with --schema_name=sometenant
```

### DIFF
--- a/django_tenants/management/commands/__init__.py
+++ b/django_tenants/management/commands/__init__.py
@@ -16,7 +16,7 @@ class BaseTenantCommand(BaseCommand):
 
     # schema name can be overridden when inheriting from BaseTenantCommand
     # if your command will always run on the same schema such as 'public'
-    schema_name = ''
+    schema_name = None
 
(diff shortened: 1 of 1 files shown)


django_tenants/management/commands/__init__.py
```diff
@@ -56,10 +60,12 @@ def handle(self, *args, **options):
         """
         Iterates a command over all registered schemata.
         """
-        if options['schema_name']:
+        if options['schema_name'] or self.schema_name:
+            # options schema_name can override inherited schema_name
+            schema_name = options['schema_name'] or self.schema_name
```
hexponent, line 65: Move it outside `if` to avoid code duplication

tomturner:
@granteagon Thanks

Issue #162, opened 2018-03-12
Management command createcachetable not working


guidezpl:
The same issue raised here https://github.com/bernardopires/django-tenant-schemas/issues/348 occurs with django-tenants.

Guo-Zhang:
This problem still occurs right now.

cs-satish-mishra:
@tomturner  The problem is still there 
@guidezpl @Guo-Zhang  any workaround you could find?

PR #994, opened 2023-11-05
Update routers.py

hho6643:
check for django_cache app (Django's DatabaseCache backend) and return "True" if it's there, fix for closed issue #162


atodorov (approved):
LGTM.

@tomturner WDYT ?

Issue #561, opened 2021-04-05
Clone tenant fails


JakeMana:
When using TENANT_CREATION_FAKES_MIGRATIONS my tenant creation fails with errors:
Does anyone have an idea how to fix it?
```
syntax error at or near "5"
LINE 695: ...one_schema(text, text, boolean, boolean) OWNER TO 5kv8307nvs...
```
![Screen Shot 2021-04-06 at 01 30 55](https://user-images.githubusercontent.com/14343741/113638714-c8daf700-9677-11eb-9e72-be1059e77892.png)

tomturner:
What permission have you got to the PG DB as it adds a function

JakeMana:
it was on Digital Ocean DB cluster so it was not an admin role.
TBH Im no sure what lvl of perms Digital Ocean gives you :/.

tomturner:
This should be fixed. The failure was an unquoted role name in generated DDL:

```
syntax error at or near "5"
LINE 695: ...one_schema(text, text, boolean, boolean) OWNER TO 5kv8307nvs...
```

`5kv8307nvs…` starts with a digit, so it has to be quoted to be a valid identifier — exactly the class of bug in #1189. The `ALTER FUNCTION … OWNER TO` in question now quotes the role, and #1239 quotes the three remaining unquoted role interpolations inside `clone_schema` (`CREATE SCHEMA … AUTHORIZATION`, `ALTER SEQUENCE … OWNER TO`, and the `ALTER TABLE … OWNER TO` in `pg_get_tabledef`), with a regression test that clones a schema owned by a role whose name needs quoting.

Closing as fixed. If you still see this on 3.12.0 or later, please reopen with the role name and the failing statement.

Issue #493, opened 2020-11-07
Fake migrations and clone tenant fails


Tekano985:
Hi, I've been trying for a couple of times to create tenants using fake migrations but I keep getting the same error:

The above exception (column "max_value" does not exist LINE 1: SELECT last_value, max_value, start_value, increment_by, min... ^ HINT: Perhaps you meant to reference the column "django_content_type_id_seq.last_value". QUERY: SELECT last_value, max_value, start_value, increment_by, min_value, cache_value, log_cnt, is_cycled, is_called FROM "test3_1604787259".django_content_type_id_seq; CONTEXT: PL/pgSQL function clone_schema(text,text,boolean) line 65 at EXECUTE ) was the direct cause of the following exception. 

How can I fix this thanks

tomturner:
Closing as part of a backlog tidy-up — not because the report was wrong, but because it can no longer be acted on as it stands.

This has been inactive for some years and a great deal has changed underneath it: `clone_schema` has been replaced wholesale from upstream pg-clone-schema, psycopg3 is now the default driver, and the supported range is Django 4.2 / 5.2 / 6.0 on PostgreSQL 14–17. There's a fair chance this no longer reproduces.

If it still happens on **3.12.0**, please reopen with the django-tenants and Django versions, the PostgreSQL version, and the steps to reproduce — with that I can actually chase it down.

Thanks for reporting it.

PR #997, opened 2023-11-16
clone: update from pg-clone-schema

rissson:
Upgrades the clone_schema function to https://github.com/denishpatel/pg-clone-schema/commit/073922e2fffe5c65c4b948f244a7caaad51429bd, which has a bunch of fixes and improvements since it's last been updated. This may fix a couple of issues that are still opened (#561, #493).

With the update, one may now choose to copy only the database structure (tables, indices, etc.) or copy the data along with it. There are also a number of other settings that can be supported (see pg-clone-schema README), but I chose not to include those in this PR.

Last change was to always apply the clone_schema function on the database, so that it gets updated if it ever gets changed in the future.


rissson:
Yep, some `dest_schema` references missed a `quote_ident` invocation. Should be fixed (I ran the test with pg11 only)

io-ma:
Hi everyone!
I would kill for this to get merged :-)

rissson:
We've been using this in https://github.com/goauthentik/authentik/pull/7590, released for about a month now and haven't encountered any particular issue. The fact that all CI jobs are failing probably indicate that I broke something in the CI with those changes, I'll have to take a closer look.

hassaanalansary:
Hey @tomturner @rissson! Just wondering what’s needed to get this PR across the finish line. Let me know if I can give it a little nudge! 🚀

hassaanalansary:
Thank you for  your response, I am still having some problems running tests using docker
however I think I have another solution for this issue,
how about we make this modular by introducing something like `TENANT_SCHEMA_CLONE_CLASS` this can enable the developers to create/use their own cloning mechanism, and this fix can distributed alone
I can work on that if you agree on this approach
    On Thursday, March 20, 2025 at 09:48:20 AM GMT+2, rissson ***@***.***> wrote:  
 
  
@hassaanalansary we need to figure out why some of the tests won't pass. As far as I remember they would pass on my machine (before the recent rebase that is), but not in CI. Since the maintainer has to re-approve the CI runs every time, it's a bit of a hassle for everyone to figure this out. Not sure how we can make that process easier for everyone.

—
Reply to this email directly, view it on GitHub, or unsubscribe.
You are receiving this because you were mentioned.Message ID: ***@***.***>
rissson left a comment (django-tenants/django-tenants#997)
@hassaanalansary we need to figure out why some of the tests won't pass. As far as I remember they would pass on my machine (before the recent rebase that is), but not in CI. Since the maintainer has to re-approve the CI runs every time, it's a bit of a hassle for everyone to figure this out. Not sure how we can make that process easier for everyone.

rissson:
Alright, I've updated clone_schema to the latest version again, and the tests pass locally on postgres 17.

> how about we make this modular by introducing something like `TENANT_SCHEMA_CLONE_CLASS` this can enable the developers to create/use their own cloning mechanism, and this fix can distributed alone

this is a good idea, but out of scope for that PR

PR #1000, opened 2023-12-04
Add Django 5.0 and Python 3.12, remove 3.1, 4.0 and Python 3.8

riconnon:
Update the dependencies to include the the latest Django release
Add it, and Python 3.12, to the test runs and remove some old (unsupported) versions from the test matrix.

### DIFF
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -60,17 +60,19 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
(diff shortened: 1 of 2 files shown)


tomturner:
Thanks @riconnon

tartieret:
@riconnon just curious but why are you removing Django 3.2 ? it's still a current LTS version for Django until April 1st 2024: https://endoflife.date/django
Wouldn't it make sense to support at least the current LTS versions, as a good practice?

PR #1006, opened 2024-01-08
Return tenant delete result

mikicz:
The returned value from `.delete()` is quite useful for showing what related objects were deleted, so I think it would be nice to return it from `TenantMixin` as well!


tomturner:
@mikicz Thanks

PR #1015, opened 2024-01-13
Use psycopg 3 instead of 2 for testing

atodorov:
NOTE: not sure if the number of queries has changed, need to take a deeper look into it. 

Also found this bit of useful information: https://learndjango.com/tutorials/psycopg3-binary-and-django-42-installation-quick-t

WARNING: this commit forces pip to use requirements.txt in CI 

WARNING: remove Django from requirements.txt b/c version is explicitly configured in CI. For local development we can use whatever version we want (will usually be the latest anyway).

### DIFF
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -24,14 +24,12 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
-    - name: Install Django
-      run: |
-        # always use latest Django b/c the point here is to test PostgreSQL compatibility
-        pip install Django
-
     - name: Install Python dependencies
       run: |
-        pip install psycopg2 coverage setuptools
+        pip install -r requirements.txt
+        pip install setuptools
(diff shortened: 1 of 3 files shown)


hexponent:
@atodorov I think we should still have Django in requirements.txt specified as Django>=4.2, because Django <= 4.1 doesn't support psycopg3

cclauss:
Given #1064, this seems to be done.
* #1064

atodorov:
> Given #1064, this seems to be done.
> 
>     * [Update psycopg requirement from <3.2,>=3.1 to >=3.2.1,<3.3 #1064](https://github.com/django-tenants/django-tenants/pull/1064)
> 

Nope, this is not done. The change referenced above only reflects what gets installed locally. The CI jobs unfortunately still specify `psycopg2` explicitly instead of using the requirements.txt file (see any recent CI job logs).

PR #1020, opened 2024-01-16
Add numeric tenant selection to tenant_command

bartonip:
Often I have something on the clipboard I need to paste into the CLI and needing to manually copy and paste the schema name into the `Enter Tenant Schema ('?' to list schemas):` prompt is frustrating.

I've added functionality that prepends the current tenant schema listing with an index, so that one can just simply type the number of the tenant they wish to execute the command on and it will do so.

It looks like this:

```
Enter Tenant Schema ('?' to list schemas): ?
1) public - public.example.com
2) apple_1659236828 - apple.example.com
3) orange_1659735437 - orange.example.com
Enter Tenant Schema ('?' to list schemas): 2
```

### DIFF
--- a/django_tenants/management/commands/__init__.py
+++ b/django_tenants/management/commands/__init__.py
@@ -95,12 +95,25 @@ https://django-tenants.readthedocs.org/en/latest/use.html#creating-a-tenant""")
             while True:
                 tenant_schema = input("Enter Tenant Schema ('?' to list schemas): ")
                 if tenant_schema == '?':
-                    print('\n'.join(["%s - %s" % (t.schema_name, t.get_primary_domain()) for t in all_tenants]))
+                    print('\n'.join(["%i) %s - %s" % (t.schema_name, t.get_primary_domain()) for i, t in enumerate(all_tenants, start=1)]))
                 else:
                     break
 
-        if tenant_schema not in [t.schema_name for t in all_tenants]:
+        tenant_schema_index = 0
+        tenant_schema_index_is_valid = False
+
+        try:
+            tenant_schema_index = int(tenant_schema) - 1
+        except ValueError:
+            pass
+        else:
+            tenant_schema_index_is_valid = 0 < tenant_schema_index <= len(all_tenants) 
+
+        if tenant_schema not in [t.schema_name for t in all_tenants] and not tenant_schema_index_is_valid:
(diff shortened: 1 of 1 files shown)


django_tenants/management/commands/__init__.py
```diff
@@ -95,13 +95,26 @@ def get_tenant_from_options_or_interactive(self, **options):
             while True:
                 tenant_schema = input("Enter Tenant Schema ('?' to list schemas): ")
                 if tenant_schema == '?':
-                    print('\n'.join(["%s - %s" % (t.schema_name, t.get_primary_domain()) for t in all_tenants]))
+                    print('\n'.join(["%i) %s - %s" % (i, t.schema_name, t.get_primary_domain()) for i, t in enumerate(all_tenants, start=1)]))
```
atodorov, line 98: why not start from zero and save a few lines trying to readjust the index ?

django_tenants/management/commands/__init__.py
```diff
@@ -95,13 +95,26 @@ def get_tenant_from_options_or_interactive(self, **options):
...
+        except ValueError:
+            pass
+        else:
+            tenant_schema_index_is_valid = 0 < tenant_schema_index <= len(all_tenants) 
+
+        if tenant_schema not in [t.schema_name for t in all_tenants] and not tenant_schema_index_is_valid:
             raise CommandError("Invalid tenant schema, '%s'" % (tenant_schema,))
 
+        if tenant_schema_index_is_valid:
+            return all_tenants[tenant_schema_index]
```
atodorov, line 116: I'm quite certain what this line is supposed to do in contrast with L118. Looks like we could just return the tenant model without querying it again.

django_tenants/management/commands/__init__.py
```diff
@@ -95,13 +95,26 @@ def get_tenant_from_options_or_interactive(self, **options):
...
-        if tenant_schema not in [t.schema_name for t in all_tenants]:
+        tenant_schema_index = 0
+        tenant_schema_index_is_valid = False
+
+        try:
+            tenant_schema_index = int(tenant_schema) - 1
+        except ValueError:
+            pass
+        else:
+            tenant_schema_index_is_valid = 0 < tenant_schema_index <= len(all_tenants)
```
atodorov, line 110: Perhaps just raise an expception is the index isn't valid at all.

atodorov (commented):
Added a few comments

Issue #1005, opened 2024-01-07
AttributeError: 'FakeTenant' object has no attribute 'domain_subfolder'


physicsb0y:
I have done everything as per the documentation for implementing sub-folder support, but when I created private tenant and go to the url, I got this error:

Exception inside application: 'FakeTenant' object has no attribute 'domain_subfolder'
Traceback (most recent call last):
  File "/media/sagardhungana/Sagar Files/ChatAppProject/myenv/lib/python3.11/site-packages/asgiref/sync.py", line 486, in thread_handler
    raise exc_info[1]
  File "/media/sagardhungana/Sagar Files/ChatAppProject/myenv/lib/python3.11/site-packages/django/core/handlers/exception.py", line 42, in inner
    response = await get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/media/sagardhungana/Sagar Files/ChatAppProject/myenv/lib/python3.11/site-packages/django/core/handlers/base.py", line 235, in _get_response_async
    callback, callback_args, callback_kwargs = self.resolve_request(request)
                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/media/sagardhungana/Sagar Files/ChatAppProject/myenv/lib/python3.11/site-packages/django/core/handlers/base.py", line 313, in resolve_request
    resolver_match = resolver.resolve(request.path_info)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/media/sagardhungana/Sagar Files/ChatAppProject/myenv/lib/python3.11/site-packages/django/urls/resolvers.py", line 665, in resolve
    sub_match = pattern.resolve(new_path)
                ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/media/sagardhungana/Sagar Files/ChatAppProject/myenv/lib/python3.11/site-packages/django/urls/resolvers.py", line 660, in resolve
    match = self.pattern.match(path)
            ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/media/sagardhungana/Sagar Files/ChatAppProject/myenv/lib/python3.11/site-packages/django_tenants/urlresolvers.py", line 50, in match
    tenant_prefix = self.tenant_prefix
                    ^^^^^^^^^^^^^^^^^^
[...]

PR #1031, opened 2024-02-17
Fix error upon creating tenant from public schema

BasVerlooy:
I faced the same issue as in https://github.com/django-tenants/django-tenants/issues/1005 and this fixed it for me. So I'm hoping we can get this merged so I can also use this in my production instance. 

For me this issue arose when I tried to create a new Tenant from the admin, via a custom TenantRequest model and having an action in the admin where I can either reject or accept such requests. It would create the tenant successfully, but when rendering the response it would fail with the error mentioned in issue #1005

### DIFF
--- a/django_tenants/urlresolvers.py
+++ b/django_tenants/urlresolvers.py
@@ -8,7 +8,10 @@ from django.utils.functional import lazy
 from django_tenants.utils import (
     get_tenant_domain_model,
     get_subfolder_prefix,
-    clean_tenant_url, has_multi_type_tenants, get_tenant_types,
+    clean_tenant_url,
+    has_multi_type_tenants,
+    get_tenant_types,
+    get_public_schema_name,
 )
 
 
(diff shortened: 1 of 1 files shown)


BasVerlooy:
@atodorov pls review/check

atodorov (changes requested):
@BasVerlooy, @physicsb0y what are the steps to reproduce this problem ?

The reported exception `Exception inside application: 'FakeTenant' object has no attribute 'domain_subfolder'` refers to a class from `base.py` while inside of `urlresolvers.py` that should point to the actual model class.  I am wondering why it isn't. 

Once we know the steps to reproduce we'll also need a test (or improve existing ones) because this seems critical enough to warrant testing.

physicsb0y:
@atodorov You can reproduce this error when using this package with daphne server in Sub-folder Support

hexponent:
@atodorov FYI
I just found another PR #822 that claims to fix the same issue
Probably would be good to learn all the proposals before making final decision, since they are very different

Issue #1058, opened 2024-06-18
Management command delete_tenant not dropping  the schema 


jahazieldom:
The management command `delete_tenant` it's not deleting the schema, when responding with "yes" it doesn't show the output "Deleting schemaname", but if I enter anything other than _"no"_ or _"yes"_, it gives me _"Please answer yes or no:"_ and when i answer _"yes"_ on that second confirm it works like expected, it's weird, like the first commit is always a "no" answer.

At first I thought it might be the shell I was using (I use zsh), but I switched to bash and it doesn't work either.

## Not working
![image](https://github.com/django-tenants/django-tenants/assets/3392725/7b609f92-6143-4455-8425-9dc404dd5f84)

## Working
![image](https://github.com/django-tenants/django-tenants/assets/3392725/d0108c2c-aabe-4b88-b3c6-74a9d02e2fd4)

PR #1059, opened 2024-06-19
#1058 fix management_command delete_tenant interactive

jahazieldom:


tomturner:
@jahazieldom Thank you for this I have now done a fix for this

Issue #1030, opened 2024-02-14
I get this error using subfoldermiddleware 


williank:
Page not found (404)
Request Method:	GET
Request URL:	http://localhost/clients/tenant2
Using the URLconf defined in app.urls_tenants_dynamically_tenant_prefixed, Django tried these URL patterns, in this order:

/

ROOT_URLCONF = 'app.urls_tenants'
PUBLIC_SCHEMA_URLCONF = 'app.urls'

I put in my settings

McDonnellJoseph:
@devmonkey22 Did #1067 work for you ?

devmonkey22:
> @devmonkey22 Did #1067 work for you ?

It did @McDonnellJoseph !  Thank you!  I didn't merge/build with your PR code directly in my project, but used the same `get_subfolder_urlconf` function body as you provided in my project for my custom tenant middleware (subclassed `TenantMainMiddleware` with mostly same logic as `TenantSubfolderMiddleware`).  I'd assume your PR would work directly as well.  My custom subclass was needed for other purposes.

tomturner:
Closing as part of a backlog tidy-up — not because the report was wrong, but because it can no longer be acted on as it stands.

This has been inactive for some years and a great deal has changed underneath it: `clone_schema` has been replaced wholesale from upstream pg-clone-schema, psycopg3 is now the default driver, and the supported range is Django 4.2 / 5.2 / 6.0 on PostgreSQL 14–17. There's a fair chance this no longer reproduces.

If it still happens on **3.12.0**, please reopen with the django-tenants and Django versions, the PostgreSQL version, and the steps to reproduce — with that I can actually chase it down.

Thanks for reporting it.

PR #1067, opened 2024-07-15
generate per-tenant urlconf in sub-folder middleware see #822

McDonnellJoseph:
Introduces the same changes as in #822 and adapts existing `url_resolvers` tests.

### DIFF
--- a/django_tenants/tests/test_urlresolvers.py
+++ b/django_tenants/tests/test_urlresolvers.py
@@ -1,11 +1,9 @@
-import sys
-from importlib import import_module
-
 from django.conf import settings
 from django.urls import reverse
 
 from django_tenants.tests.testcases import BaseTestCase
-from django_tenants.urlresolvers import TenantPrefixPattern, get_subfolder_urlconf
+from django_tenants.urlresolvers import get_subfolder_urlconf
+
 from django_tenants.utils import get_tenant_model, get_tenant_domain_model
 
 
@@ -27,11 +25,8 @@ class URLResolversTestCase(BaseTestCase):
             """
             Reverses `name` in the urlconf returned from `tenant`.
             """
-
-            urlconf_path = get_subfolder_urlconf(tenant)
-            urlconf = import_module(urlconf_path)
+            urlconf = get_subfolder_urlconf(tenant)
             reverse_response = reverse(name, urlconf=urlconf)
-            del sys.modules[urlconf_path]  # required to simulate new thread next time
             return reverse_response
 
         cls.reverser = reverser_func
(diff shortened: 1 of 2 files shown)


devmonkey22:
To bring the conversation from #1030 here, I didn't merge/build with your PR code directly in my project, but used the same `get_subfolder_urlconf` function body as you provided in my project for my custom tenant middleware (subclassed `TenantMainMiddleware` with mostly same logic as `TenantSubfolderMiddleware`). I'd assume your PR would work directly as well. My custom subclass was needed for other purposes.

glendza:
Thank you for this, monkey-patched the `get_subfolder_urlconf` and it started working after a lot of frustration. Really hope this saga officially comes to a finish soon.

McDonnellJoseph:
Hey @tomturner @atodorov I noticed you expressed interest in the previous PR and it seems to be helping several users. If you have some time I would appreciate some feedback on these new changes :smile:

mronian:
We are facing the same issue, @atodorov @tomturner would really help if you merge these changes 🙌

twentyforty:
Getting this in would help a ton.  @atodorov @tomturner could you please take a look?

sicter:
This fixed the issue I was having (brought here from #1005 ) ! If pointed in the right direction, I'd be happy to try to get this across the finish line, what's missing?

NotANormalNerd:
I have a similar problem, where a reverse for a redirect isn't properly resolved to the tenant. I couldn't figure out where it is happening and this is probably also the most likely explanation. 
My problem is that I can't reproduce it locally but I can reproduce it in our productve envs.

tomturner:
Reviewed this properly and I like it — baking the subfolder into a static `path()` removes the per-resolution `Domain` query that `TenantPrefixPattern.tenant_prefix` was doing on every reverse, which is the right fix (and makes #768's caching workaround unnecessary). Rebased on current master the full suite passes: 98 tests, OK.

I'm holding it for **4.0** rather than 3.11.0, because it removes public API:

- `tenant_patterns()`
- `TenantPrefixPattern`
- `get_dynamic_tenant_prefixed_urlconf()`
- and `get_subfolder_urlconf()` changes its return type from a module path `str` to a module object

Nothing in the docs or the package outside the tests uses them, but `tenant_patterns` is the documented-by-example way people wire up subfolder tenants in their own root urlconf, so removing it is a breaking change that shouldn't land in a minor release.

Nothing needed from you for now — leaving it open and targeting 4.0. When we get there the only open question is whether to keep `tenant_patterns`/`TenantPrefixPattern` as deprecation shims for a release, or drop them outright with a note in the release notes.

PR #1081, opened 2024-09-03
Handling Sequence Issues When Cloning Schemas in PostgreSQL

begyy:
Problem Statement
When working with PostgreSQL, we encountered an issue related to sequence numbers in a table after deleting specific records. The problem manifested when cloning a schema, leading to potential conflicts and errors when inserting new records into the cloned table.

Scenario
We had a table product in the schema cola, which originally contained the following records:

product_id=1
product_id=2
product_id=3
product_id=4
After performing some operations, we deleted the records with product_id=1, product_id=3, and product_id=4, leaving only the record with product_id=2.

The sequence associated with the product_id field continued from the last highest number, resulting in the next product_id being 5 upon insertion of a new record. However, during schema cloning, the sequence reset based on the current row count rather than maintaining the correct sequence. This led to errors when new records were inserted because the sequence did not account for the missing IDs.


begyy:
Screenshot 2024-09-03 at 12 49 27

tomturner:
You have done a lot of changes to clone Cloning was based of https://github.com/denishpatel/pg-clone-schema/. would this still be the case?

begyy:
> You have done a lot of changes to clone Cloning was based of https://github.com/denishpatel/pg-clone-schema/. would this still be the case?

Hello, yes I took the code from there and changed it a little :)

very strange I tested postgres 13, 14, 15, 16 everything was fine but in the github test it doesn't work :((

PR #1097, opened 2024-10-23
Add Support for Managing Domains with Commands

rpy-sclark:
We're unable to access the database directly and need to use management commands for domains.

These commands mirror functionality of the create_tenant and delete_tenant commands but for domains.

I welcome all feedback!


tomturner:
@rpy-sclark Thanks

PR #1099, opened 2024-11-05
Send a signal before migrations run

mikicz:
Django itself has signals pre/post migrations (https://github.com/django/django/blob/main/django/db/models/signals.py#L53), so this introduces the pre signal for django-tenants. We already have `schema_migrated` which is the post signal


atodorov (approved):
LGTM, @tomturner WDYT ?

tomturner:
@mikicz thanks

PR #1106, opened 2024-12-02
Support database alias in schema_context() & tenant_context()

atodorov:
similar to what was already introduced in d02a4ee0. To use it:

```
with schema_context("my_schema", database="replica_db"):
    pass
```


tomturner:
@atodorov thanks

PR #1109, opened 2024-12-11
Feature/make host compose ports configurable

ethagnawl:
This PR introduces .env.sample which, when copied to the ignored .env, can be used to provide custom host ports to the Docker Compose config.


MikeSoft:
is not neccesary the `env_file` in the docker-compose?

ethagnawl:
I believe .env is read in implicitly. See: https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#additional-information

PR #1113, opened 2024-12-24
Add LRU cache to hostname extraction for improved performance

MikeSoft:
Add LRU cache to hostname_from_request method

Implements LRU (Least Recently Used) caching with a maximum size of 128 entries for the hostname_from_request method. This optimization:

- Improves performance by avoiding repeated processing of the same hostnames
- Reduces request processing overhead by caching frequently accessed results 
- Maintains memory bounds by setting maxsize=128

Since hostname_from_request is called on every request to extract and normalize the hostname, the cache will particularly help in high-traffic scenarios.

### DIFF
--- a/django_tenants/middleware/main.py
+++ b/django_tenants/middleware/main.py
@@ -1,3 +1,4 @@
+from functools import lru_cache
 from django.conf import settings
 from django.core.exceptions import DisallowedHost
 from django.db import connection
@@ -19,6 +20,7 @@ class TenantMainMiddleware(MiddlewareMixin):
     """
 
     @staticmethod
+    @lru_cache(maxsize=128)
     def hostname_from_request(request):
         """ Extracts hostname from request. Used for custom requests filtering.
             By default removes the request's port and common prefixes.


MikeSoft:
when read this: 😂image

xpund:
I previously implemented a similar caching solution for my project, and it has been working successfully. You can find the implementation here: [django-tenants PR #1121](https://github.com/django-tenants/django-tenants/pull/1121).

In your implementation, I noticed a couple of areas for improvement:

Cache Invalidation: It's important to provide a mechanism for clearing or invalidating the cache when tenant data changes. Without this, the cache might serve stale data, which could lead to unexpected behavior.

typed=True in @lru_cache: Using typed=True ensures that the cache differentiates between variations in data types (e.g., strings vs. bytes) for the same hostname. This prevents potential issues with cache collisions or incorrect data retrieval.

Additionally, you might want to consider adding tests for scenarios where the cache is invalidated or when typed=True behavior affects the caching logic. These tests can ensure that the implementation handles edge cases effectively.

Let me know if you'd like any further insights or examples!

MikeSoft:
> I previously implemented a similar caching solution for my project, and it has been working successfully. You can find the implementation here: [django-tenants PR #1121](https://github.com/django-tenants/django-tenants/pull/1121).
> 
> In your implementation, I noticed a couple of areas for improvement:
> 
> Cache Invalidation: It's important to provide a mechanism for clearing or invalidating the cache when tenant data changes. Without this, the cache might serve stale data, which could lead to unexpected behavior.
> 
> typed=True in @lru_cache: Using typed=True ensures that the cache differentiates between variations in data types (e.g., strings vs. bytes) for the same hostname. This prevents potential issues with cache collisions or incorrect data retrieval.
> 
> Additionally, you might want to consider adding tests for scenarios where the cache is invalidated or when typed=True behavior affects the caching logic. These tests can ensure that the implementation handles edge cases effectively.

Your approach is different,  I only cache get tenant from url, you are caching getting the tenant from database. 
Both should be together :)

xpund:
Your cache must be cleared in any case when domains or tenants are modified.
However, as far as I know, `lru_cache` cannot be cleared or updated by key.
Therefore, the entire cache will need to be cleared whenever changes occur.

xpund:
Oh, sorry 
I glanced at your code the first time and misunderstood your solution.

Yes, your approach is indeed appropriate; there's no point in clearing the cache. However, adding `typed=True` does make sense; otherwise, it will return the same data for any parameter.

I'm also concerned about the `get_host()` method call. It seems more resource-intensive compared to `split()`.

django_tenants/middleware/main.py
```diff
@@ -19,23 +20,45 @@ class TenantMainMiddleware(MiddlewareMixin):
     """
 
     @staticmethod
-    def hostname_from_request(request):
-        """ Extracts hostname from request. Used for custom requests filtering.
-            By default removes the request's port and common prefixes.
+    @lru_cache(maxsize=128)
+    def hostname_from_request_host(request_host):
```
foarsitter, line 24: Factually your are right, the parameter name and method name do not cover what is going on and could be improved. But, that does not justifies introducing a backwards incompatible change.

django_tenants/middleware/main.py
```diff
@@ -19,23 +20,45 @@ class TenantMainMiddleware(MiddlewareMixin):
...
 
     def get_tenant(self, domain_model, hostname):
+        """
+        Retrieves the tenant associated with the given hostname.
+
+        :param domain_model: The domain model to query.
+        :param hostname: The hostname to look up.
+        :return: The tenant associated with the hostname.
+        """
         domain = domain_model.objects.select_related('tenant').get(domain=hostname)
```
foarsitter, line 45: Is this docstring more descriptive then `get_tenant`?  It has nothing to do with your actual change so I would have left it unchanged.

django_tenants/middleware/main.py
```diff
@@ -19,23 +20,45 @@ class TenantMainMiddleware(MiddlewareMixin):
...
-        # the tenant metadata is stored.
+        """
+        Processes the incoming request to set the appropriate tenant schema.
+
+        Connection needs first to be at the public schema, as this is where
+        the tenant metadata is stored.
+
+        :param request: The incoming HTTP request.
+        :return: An HTTP response if the tenant is not found, otherwise None.
+        """
```
foarsitter, line 57: You are moving an important comment inside a docstring somewhere in the middle so it will be easily missed. Is that a good idea? 

If you want to improve the comment I would suggest the following:  
```
# Tenant metadata is stored in the public schema
connection.set_schema_to_public()
```
This tells us WHY were are calling  `set_schema_to_public`

foarsitter (commented):
I appreciate that you've applied the Boy Scout Rule and added some documentation, that’s a practice I generally encourage. However, in this specific instance, it is distracting. 

My second point of feedback would be to not introduce breaking changes unless there isn't another way.

Lastly, I'm wondering what the performance benefits are. Did you run some benchmarks or did it solve a real world problemen? A high traffic scenario would need a much larger cache size if they have a lot of tenants I suppose.

PR #1121, opened 2025-01-13
Add Tenant Caching with Django Cache

xpund:
This PR introduces a caching mechanism for tenant management to improve performance by reducing database queries for tenant data. A new constant, TENANT_CACHE_ENABLE, is added to toggle caching functionality, and tenant data is stored in a shared cache (TENANTS_CACHE_DATA).

**Key Changes:**

**1. Added caching logic in get_tenant method:**
When caching is enabled, tenant data is retrieved from the cache if available.
If not found, the data is fetched from the database, stored in the cache, and returned.

**2. Introduced get_cached_tenant method:**
Handles retrieving tenant data from the cache or database while ensuring cache consistency.

**3. Improved performance:**
Reduces repeated database queries for the same tenant, particularly in high-traffic environments.

**How to Use**
Enable caching by setting the following constant in your settings:

```
TENANT_CACHE_ENABLE = True
```

The cache is automatically used for tenant lookups when enabled.

### DIFF
--- a/django_tenants/middleware/main.py
+++ b/django_tenants/middleware/main.py
@@ -12,6 +12,7 @@ from django_tenants.utils import remove_www, get_public_schema_name, get_tenant_
 TENANTS_CACHE_DATA = {}
 TENANT_CACHE_ENABLE = getattr(settings, 'TENANT_CACHE_ENABLE', False)
 
+
 class TenantMainMiddleware(MiddlewareMixin):
     TENANT_NOT_FOUND_EXCEPTION = Http404
     """
@@ -26,13 +27,6 @@ class TenantMainMiddleware(MiddlewareMixin):
             By default removes the request's port and common prefixes.
         """
         return remove_www(request.get_host().split(':')[0])
-
-    def get_tenant(self, domain_model, hostname):
-        if TENANT_CACHE_ENABLE is True:
-            self.get_tenant_cache(domain_model, hostname)
-        else:
-            domain = domain_model.objects.select_related('tenant').get(domain=hostname)
-            return domain.tenant
 
     @staticmethod
     def get_cached_tenant(domain_model, hostname):
@@ -53,6 +47,13 @@ class TenantMainMiddleware(MiddlewareMixin):
             else:
                 TENANTS_CACHE_DATA[hostname] = domain.tenant
                 return domain.tenant
+
+    def get_tenant(self, domain_model, hostname):
+        if TENANT_CACHE_ENABLE is True:
+            return self.get_cached_tenant(domain_model, hostname)
+        else:
+            domain = domain_model.objects.select_related('tenant').get(domain=hostname)
+            return domain.tenant
 
     def process_request(self, request):
         # Connection needs first to be at the public schema, as this is where
--- a/django_tenants/middleware/main.py
+++ b/django_tenants/middleware/main.py
@@ -7,8 +7,10 @@ from django.utils.module_loading import import_string
 from django.utils.deprecation import MiddlewareMixin
 
 from django_tenants.utils import remove_www, get_public_schema_name, get_tenant_types, \
-    has_multi_type_tenants, get_tenant_domain_model, get_public_schema_urlconf
+    has_multi_type_tenants, get_tenant_domain_model, get_public_schema_urlconf, get_tenant_model
 
+TENANTS_CACHE_DATA = {}
+TENANT_CACHE_ENABLE = getattr(settings, 'TENANT_CACHE_ENABLE', False)
 
 class TenantMainMiddleware(MiddlewareMixin):
     TENANT_NOT_FOUND_EXCEPTION = Http404
@@ -26,8 +28,31 @@ class TenantMainMiddleware(MiddlewareMixin):
         return remove_www(request.get_host().split(':')[0])
 
     def get_tenant(self, domain_model, hostname):
-        domain = domain_model.objects.select_related('tenant').get(domain=hostname)
-        return domain.tenant
+        if TENANT_CACHE_ENABLE is True:
+            self.get_tenant_cache(domain_model, hostname)
+        else:
+            domain = domain_model.objects.select_related('tenant').get(domain=hostname)
+            return domain.tenant
+
+    @staticmethod
+    def get_cached_tenant(domain_model, hostname):
+        """Get cached tenant."""
+        if hostname in TENANTS_CACHE_DATA:
+            tenant = TENANTS_CACHE_DATA[hostname]
+            client_model = get_tenant_model()
+            if isinstance(tenant, client_model):
+                return tenant
+            else:
+                raise domain_model.DoesNotExist()
+        else:
+            try:
+                domain = domain_model.objects.select_related('tenant').get(domain=hostname)
+            except domain_model.DoesNotExist:
+                TENANTS_CACHE_DATA[hostname] = None
+                raise domain_model.DoesNotExist()
+            else:
+                TENANTS_CACHE_DATA[hostname] = domain.tenant
+                return domain.tenant
 
     def process_request(self, request):
         # Connection needs first to be at the public schema, as this is where
--- a/django_tenants/middleware/main.py
+++ b/django_tenants/middleware/main.py
@@ -23,8 +23,8 @@ class TenantMainMiddleware(MiddlewareMixin):
 
     @staticmethod
     def hostname_from_request(request):
-        """ Extracts hostname from request. Used for custom requests filtering.
-            By default removes the request's port and common prefixes.
+        """Extracts hostname from request. Used for custom requests filtering.
+        By default removes the request's port and common prefixes.
         """
         return remove_www(request.get_host().split(':')[0])
 
@@ -64,6 +64,7 @@ class TenantMainMiddleware(MiddlewareMixin):
             hostname = self.hostname_from_request(request)
         except DisallowedHost:
             from django.http import HttpResponseNotFound
+
             return HttpResponseNotFound()
 
         domain_model = get_tenant_domain_model()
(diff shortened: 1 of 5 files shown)


MikeSoft:
I implemented a custom cached middleware for pesonal project and i created a new custom middleware to avoid this: 
`if TENANT_CACHE_ENABLE is True:` for each request. 

What do you think?
Anyway, I liked your PR, good job!

xpund:
> I implemented a custom cached middleware for pesonal project and i created a new custom middleware to avoid this: `if TENANT_CACHE_ENABLE is True:` for each request.
> 
> What do you think? Anyway, I liked your PR, good job!

I think it's better to exclude `TENANT_CACHE_ENABLE` from settings and always use Django's cache. This approach is commonly used in Django's sites framework.

But since it's an open-source project, some people might not want to use this approach.
The `if TENANT_CACHE_ENABLE is True` check is lightweight and better than making a database query every time.

MikeSoft:
> > I implemented a custom cached middleware for pesonal project and i created a new custom middleware to avoid this: `if TENANT_CACHE_ENABLE is True:` for each request.
> > What do you think? Anyway, I liked your PR, good job!
> 
> I think it's better to exclude `TENANT_CACHE_ENABLE` from settings and always use Django's cache. This approach is commonly used in Django's sites framework.
> 
> But since it's an open-source project, some people might not want to use this approach. The `if TENANT_CACHE_ENABLE is True` check is lightweight and better than making a database query every time.

I agree with you that open-source projects should be different. 

`    def get_tenant(self, hostname: str):
        cache_key = f"tenant:domain:{hostname}"
        cached_tenant = cache.get(cache_key)

        if cached_tenant is not None:
            return cached_tenant

        try:
            domain = (
                Domain.objects.select_related("tenant")
                .only(
                    "tenant__activated",
                    "tenant__id",
                    "tenant__schema_name",
                    "tenant__name",
                )
                .get(domain=hostname, activated=True, tenant__activated=True)
            )
            tenant = domain.tenant
            cache.set(cache_key, tenant, self.cache_timeout)
            return tenant
        except Exception as e:
            logger.exception(
                "cached_tenant.TenantMainMiddleware.get_tenant",
                exc_info=True,
                extra={"error": str(e), "hostname": hostname},
            )
            return None`

I know that the cache key is not the best. 
What do you think about using the django.cache framework?

xpund:
> > > I implemented a custom cached middleware for pesonal project and i created a new custom middleware to avoid this: `if TENANT_CACHE_ENABLE is True:` for each request.
> > > What do you think? Anyway, I liked your PR, good job!
> > 
> > 
> > I think it's better to exclude `TENANT_CACHE_ENABLE` from settings and always use Django's cache. This approach is commonly used in Django's sites framework.
> > But since it's an open-source project, some people might not want to use this approach. The `if TENANT_CACHE_ENABLE is True` check is lightweight and better than making a database query every time.
> 
> I agree with you that open-source projects should be different.
> 
> I know that the cache key is not the best. 
> What do you think about using the django.cache framework?
> ```

django.core.cache uses various clients, and a query to a database like Redis, for example, can also create some load under high traffic, as well as introduce slight latency compared to the application's internal cache.

tomturner:
@atodorov what do you think?

django_tenants/middleware/main.py
```diff
@@ -7,7 +7,10 @@
 from django.utils.deprecation import MiddlewareMixin
 
 from django_tenants.utils import remove_www, get_public_schema_name, get_tenant_types, \
-    has_multi_type_tenants, get_tenant_domain_model, get_public_schema_urlconf
+    has_multi_type_tenants, get_tenant_domain_model, get_public_schema_urlconf, get_tenant_model
+
+TENANTS_CACHE_DATA = {}
```
atodorov, line 12: Why not use the django.cache mechanism instead ? 

From what I can tell this may not be thread safe!

django_tenants/middleware/main.py
```diff
@@ -20,14 +23,37 @@ class TenantMainMiddleware(MiddlewareMixin):
...
         """
         return remove_www(request.get_host().split(':')[0])
 
+    @staticmethod
+    def get_cached_tenant(domain_model, hostname):
+        """Get cached tenant."""
+        if hostname in TENANTS_CACHE_DATA:
+            tenant = TENANTS_CACHE_DATA[hostname]
+            client_model = get_tenant_model()
+            if isinstance(tenant, client_model):
```
atodorov, line 37: This check appears a bit redundant

django_tenants/middleware/main.py
```diff
@@ -20,14 +23,37 @@ class TenantMainMiddleware(MiddlewareMixin):
...
+                return tenant
+            else:
+                raise domain_model.DoesNotExist()
+        else:
+            try:
+                domain = domain_model.objects.select_related('tenant').get(domain=hostname)
+            except domain_model.DoesNotExist:
+                TENANTS_CACHE_DATA[hostname] = None
```
atodorov, line 45: Don't assign anything here aka remove this line and then you won't need the if condition on line 37, no ?

django_tenants/signals.py
```diff
@@ -50,3 +55,47 @@ def tenant_delete_callback(sender, instance, **kwargs):
...
+        return
+
+    # copy() is required
+    # (RuntimeError: dictionary changed size during iteration)
+    for hostname, tenant in TENANTS_CACHE_DATA.copy().items():
+        if tenant is None or tenant == instance:
+            try:
+                del TENANTS_CACHE_DATA[hostname]
+            except KeyError:
+                ...
```
atodorov, line 101: why not just use `pass` ? I find it a bit more clear and not everyone may be familiar with the `...` syntax.

atodorov (changes requested):
Added a few comments about the implementation + some improvement suggestions. 

However the benefits of this whole feature seems very questionable to me. See my next comment.

atodorov:
> Reduces repeated database queries for the same tenant, particularly in high-traffic environments.

I have a number of questions about how much we gain introducing this feature. 

1. How much traffic are you talking about? How many http req/sec - how many DB queries/sec  for example ? 
2. How many ms do we actually save ? Have you done any measurements and what are the numbers ? 


In particular, I feel like saving 1 DB query to fetch the tenant object isn't much of a practical gain b/c the slowest portion is actually establishing the DB connection. In the case where you have lots of short-lived requests this is worse for Postgres than for MySQL (I recently measured this for my own use case and Postgres performed nearly 40% worse than MariaDB in some scenarios, see -> https://kiwitcms.readthedocs.io/en/latest/hardware_performance.html). 

^^^ the above is to say that using the new DB connection pooling feature in Django 5.1 is most likely going to result in more performance benefits than saving 1 DB query. (this is something I have on my todo list to explore for my own environment).

django_tenants/middleware/main.py
```diff
@@ -20,14 +23,37 @@ class TenantMainMiddleware(MiddlewareMixin):
...
+                raise domain_model.DoesNotExist()
+            else:
+                TENANTS_CACHE_DATA[hostname] = domain.tenant
+                return domain.tenant
+
     def get_tenant(self, domain_model, hostname):
-        domain = domain_model.objects.select_related('tenant').get(domain=hostname)
-        return domain.tenant
```
atodorov, line 30: A final though: 

if indeed we go forward with caching why not just cache the result of this DB query using Django's existing cache backend? 

```
        domain = cache.get(hostname, domain_model.objects.select_related('tenant').get(domain=hostname))
        return domain.tenant
```

?

atodorov: FTR I see the previous comments above around using Django's existing cache framework and

* They talk about giving choice, but the current implementation gives none - it's hard-coded. 
* Django's cache framework provides different backends and you can write your own if need be. If anyone doesn't want to use Redis (for whatever reason) they can default to `django.core.cache.backends.locmem.LocMemCache` which is similar to the current implementation with a global variable in the module.
* The argument that "open source project should be different" is just crazy to me. Let's not reinvent the wheel and use the existing features provided by Django - everyone needing to use cache is most likely already familiar with Django's existing cache framework and they probably have the necessary infrastructure for that.

tomturner:
Thanks for this — cutting the per-request domain lookup is a worthwhile goal.

I'm closing it because of two problems I confirmed by testing the branch (rebased on current master, since on its original base it fails Django 5.2's `migrate_and_makemigrations_autodetector` check). The 2 tests in the PR do pass; these are things they don't cover.

**1. A tenant row change leaves a stale object in the cache.**

`delete_or_change_client()` is documented "Delete client from cache", but its guard is:

```python
if TENANT_CACHE_ENABLE is not True or not isinstance(instance, get_tenant_domain_model()):
    return
```

That's the *Domain* model, not the tenant model, so the receiver never fires for a tenant save or delete — it's dead code. (Deleting a tenant does get evicted, but via the `Domain` cascade reaching `delete_domain`, not via this function.) A plain `tenant.save()` leaves the cached instance behind:

```
value in DB     : 'CHANGED IN THE DATABASE'
value in cache  : None
cache is stale  : True
```

**2. Unknown hostnames grow the cache without bound.**

Misses are cached as `None` and never evicted. With the wildcard `ALLOWED_HOSTS` a subdomain-per-tenant deployment needs, any unknown subdomain reaches `get_tenant()` and leaves a permanent entry:

```
entries after 5000 unknown subdomains: 5000
every one a negative entry: True
```

That is unauthenticated, attacker-controlled growth in a process-lifetime dict — a memory-exhaustion vector rather than a cache.

Two smaller things worth carrying into any follow-up:

- Despite the title, this doesn't use Django's cache framework — `TENANTS_CACHE_DATA` is a module-level dict, so it is per-worker and diverges between processes, and there's no way to invalidate across them.
- `TENANT_CACHE_ENABLE` is read at import time in two modules, which is why the test has to `patch('django_tenants.middleware.main.TENANT_CACHE_ENABLE', True)` instead of just using `override_settings`. Reading it inside the function would fix that.
- The three `post_save`/`post_delete` receivers are registered for every model in the project and call `get_tenant_domain_model()` on every save, even when the feature is off.

A version built on `django.core.cache` with a TTL, no negative caching (or a bounded one), and invalidation keyed on the tenant model would be very welcome.

PR #1041, opened 2024-03-28
Add tenant attribute to request object

Dresdn:
Add the `tenant` attribute to `request` objects generated by `TenantRequestFactory`, simulating what all the middleware options do.

Closes #1040

PR #1132, opened 2025-03-07
Apply TenantMiddleware to TenantRequestFactory request

hexponent:
Although there was middleware simulation introduced in #1041, it does not solve all issues during testing.
In my code, I rely on `tenant.domain_url`, another [attribute set by middleware](https://github.com/django-tenants/django-tenants/blob/v3.7.0/django_tenants/middleware/main.py#L49), but since middleware is not activated - this attribute is `None`(default value from `TenantMixin`) instead of `str`, causing unexpected errors.

Instead of copy-pasting middleware functionality in RequestFactory little by little, here I propose that we just call middleware, especially since it is already present in RequestFactory class.


hexponent:
@atodorov please review

scur-iolus:
It is explicitly said in Django documentation that the `RequestFactory` "_does not support middleware_" ([here](https://docs.djangoproject.com/en/5.2/topics/testing/advanced/#django.test.RequestFactory)). This being said, I agree it could be relevant to make an exception for `TenantRequestFactory` and `TenantMainMiddleware`, but in that case the documentation should highlight this point.

I suspect the issue #1142, that I just reported and which is also related to `TenantRequestFactory`, was also introduced by the pull request #1041 (I haven't had the time to look for a solution yet).

django_tenants/test/client.py
```diff
@@ -14,14 +14,16 @@ def __init__(self, tenant, **defaults):
...
 
 
 class TenantRequestFactory(BaseTenantRequestFactory, RequestFactory):
-    pass
+
+    def generic(self, *args, **kwargs):
+        request = super().generic(*args, **kwargs)
+        # Process request using Tenant middleware
+        self.tm.process_request(request)
+        return request
```
atodorov, line 26: I think this could benefit from explicit documentation inside README and/or doc-strings for the `BaseTenantRequestFactory` class.

In fact this library now supports several different middleware implementations (TenantMainMiddleware, TenantSubfolderMiddleware, SuspiciousTenantMiddleware) and folks may contribute more over time. 

Yet at present TenantClient defaults to a specific middleware implementation. That should be easier to override in one's tests.

atodorov: Seeing how we also have SubfolderTenantTestCase which doesn't appear to override the middleware (from what I can see adjusting the request client is left to the developer - that's what I have in my own tests). 

This has the potential to break lots of things in unexpected ways.

atodorov (changes requested):
I like the PR in principle, however it looks like this simple change will have far reaching consequences for lots of people. Most likely this is  a breaking change that requires a major version release. 

Not sure how to resolve this though. 

Setting to "Request changes" for now so we don't merge it accidentally and make everyone's lives difficult.

PR #1133, opened 2025-03-20
Fix: cloning when TENANT_BASE_SCHEMA does not exist in the DB

hassaanalansary:
clone_schema fails silently if TENANT_BASE_SCHEMA does not exist in the DB. Thus it should fallback to creating the schema.
Possible example: first time to run test cases and it is a blank database


hassaanalansary:
kind reminder

PR #1138, opened 2025-04-03
Django 5.2

riconnon:


tomturner:
The tests have to pass before this goes in.  I have been working on!!!

.github/workflows/code.yml
```diff
@@ -51,8 +51,8 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ["3.10", "3.11", "3.12"]
-        django-version: ["==4.2.*", "==5.0.*", "==5.1.*"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
+        django-version: ["==4.2.*", "==5.0.*", "==5.1.*", "==5.2.*"]
```
atodorov, line 55: Django 5.0 is EOL since Apr 1st so IMO no need to have it in the matrix.

django_tenants/management/commands/migrate_schemas.py
```diff
@@ -14,6 +15,7 @@
 
 
 class MigrateSchemasCommand(SyncCommon):
+    autodetector = MigrationAutodetector
```
atodorov, line 18: Question: What is this and why do we need it? 

Is that related to Django 5.2 support ? 

If it isn't I'd prefer to have it as a separate change.

setup.py
```diff
@@ -53,7 +53,7 @@
         'Topic :: Software Development :: Libraries :: Python Modules'
     ],
     install_requires=[
-        'Django >=2.1,<5.2',
+        'Django >=2.1,<6.0',
```
atodorov, line 56: IMO the requirements here should be bumped to `>=4.2` since this is now the minimum officially supported version of Django and also the minimum version we test with. @tomturner WDYT ?

atodorov (changes requested):
Added a few minor comments. 

I am also interested in relaxing the install_requires dependency so I can migrate to Django 5.2 in my own code base. I can take a look at the failing tests soemtime soon.

setup.py
tomturner: @atodorov thats fine I agree.  If you don't mind looking at the tests that would be great.  I am quite busy at the moment and I will be watching the Django con talks this week.

I did a seperate branch nothing different to this in the end and I found that i needed the `autodetector = MigrationAutodetector` line.

atodorov: > @atodorov thats fine I agree. If you don't mind looking at the tests that would be great.

FTR I took a quick look b/c I have lots of failing items on my side. It appears that `_fixture_setup()` has been changed from an instance method to a class method + there are additional changes which I still don't quite understand. 

Relevant upstream ticket: https://code.djangoproject.com/ticket/35660

It's true we've been depending on an internal method for some customized testing needs and this is now broken. I will circle back to this once I have more understanding of the new behavior.

django_tenants/management/commands/migrate_schemas.py
riconnon: This is related to Django 5.2. This field was added to the migrate command in Django 5.2

setup.py
foarsitter: Somehow the app registration is meshed up, adding the registration to `_pre_setup` solves it but I don't know why. However, we can no longer rely on _pre_setup because its signature has changed, making it incompatible. Therefore, we need to either implement some Django-specific version logic or find an alternative setup method.

```python
class TenantSyncTest(BaseSyncTest):
    @classmethod
    def _pre_setup(cls):
        cls.sync_shared()
        cls.INSTALLED_APPS = cls.SHARED_APPS + cls.TENANT_APPS

        settings.SHARED_APPS = cls.SHARED_APPS
        settings.TENANT_APPS = cls.TENANT_APPS
        settings.INSTALLED_APPS = cls.INSTALLED_APPS

        cls.available_apps = cls.INSTALLED_APPS

        super()._pre_setup()
```

foarsitter: Found a solution: https://github.com/django-tenants/django-tenants/pull/1148

foarsitter:
@riconnon with https://github.com/django-tenants/django-tenants/pull/1148 merged, do you have time to finish your PR?

riconnon:
I will try and pick it up soon but I will not be offended if someone else wants to pick it up in the mean time

tomturner:
OBE by https://github.com/django-tenants/django-tenants/pull/1151

However thank you to @riconnon

PR #1144, opened 2025-05-04
Take advantage of pyproject.toml

foarsitter:
This PR replaces `setup.py` with the more modern `pyproject.toml`. 

pyproject.toml is part of the PEP 517/518 standards, providing a unified and modern way to define build system requirements and configurations.

I moved the requirements listed in requirements.txt to an optional dev group so we can install everything with a single command. Because setuptools is listed as build backend it no longer needs to be installed manually. 

Building a package can be done with the following command `python -m build --sdist`.


foarsitter:
@tomturner rebased the PR

PR #1146, opened 2025-05-08
Use literalinclude

foarsitter:
Did an experiment by replacing static examples with references to actual code by using the `literalinclude` directive. 

I think it is in improvement to use tested code as examples to prevent that the examples getting out of sync. 

As a bonus I stumbled upon `TEMPLATE_CONTEXT_PROCESSORS` which is removed and no part of `TEMPLATES`.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -13,47 +13,32 @@ You'll have to make the following modifications to your ``settings.py`` file.
 
 Your ``DATABASE_ENGINE`` setting needs to be changed to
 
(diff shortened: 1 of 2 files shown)


foarsitter:
Yes that is correct @tomturner. 

However, the directive has also options for start/end-at/after text. With these options a example can be made relative but that requires changes to the codebase. 

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude

PR #1148, opened 2025-05-12
Refactor to use class properties for settings configuration

foarsitter:
This pull request refactors the test setup by centralizing the configuration of the SHARED_APPS and TENANT_APPS properties. This change is necessary for upgrading to Django 5.2 https://github.com/django-tenants/django-tenants/pull/1138

In Django 5.2, _pre_setup() is now invoked within setupClass(). As a result, calling super().setUpClass() already triggers _pre_setup(), initializing Django with the apps hardcoded in BaseTestCase.setUpClass().

Therefore, setting the SHARED_APPS and TENANT_APPS properties after the super() call has no effect.

Additionally, I could not find a reference to cls.INSTALLED_APPS, so I have removed it.

### DIFF
--- a/django_tenants/tests/test_tenants.py
+++ b/django_tenants/tests/test_tenants.py
@@ -333,17 +333,6 @@ class BaseSyncTest(BaseTestCase):
                    'django.contrib.contenttypes', )  # 1 table
     TENANT_APPS = ('django.contrib.sessions', )
 
-    @classmethod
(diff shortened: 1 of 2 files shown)


tomturner:
Seemed ok to me so i merged it

ChrisCrossCrash:
Thanks for working on this! `django-tenants` works great with Django 5.2 now.

PR #1149, opened 2025-05-12
Formally support python 3.13

foarsitter:
Added Python 3.13 to GitHub Actions and updated the classifiers accordingly. Removed unsupported Python versions 3.8 and 3.9, as they are no longer tested.


tomturner:
I will add the 3.13 line in however i want the the following to remain until we drop support for Django 4.2

        'Programming Language :: Python :: 3.8',
        'Programming Language :: Python :: 3.9',

PR #1151, opened 2025-05-21
Django52

foarsitter:
Based on https://github.com/django-tenants/django-tenants/pull/1138 this PR adds compatibility with Django 5.2

Postgres 13 isn't longer supported and it requires a version check for fixing the `_pre_setup` change.


ChrisCrossCrash:
Thanks for your work on this!

PR #1159, opened 2025-06-09
Add configurable urlconf override for subfolder middleware see #1067

sicter:
Add ability to override subfolder tenant urlconf resolution. This enables temporary solution for devs to have a lighter weight way to apply the fix in #1067 / #822

Added a class-level `get_subfolder_urlconf_override` attribute to TenantSubfolderMiddleware that allows subclasses to customize how urlconfs are resolved for tenants in subfolder paths. When set, this override takes precedence over the default get_subfolder_urlconf function that seems to be where some issues are originating.

Example:
```python
class CustomTenantSubfolderMiddleware(TenantSubfolderMiddleware):
    @staticmethod
    def get_subfolder_urlconf(tenant):
    if has_multi_type_tenants():
        urlconf = get_tenant_types()[tenant.get_tenant_type()]["URLCONF"]
    else:
        urlconf = settings.ROOT_URLCONF

    subfolder_prefix = get_subfolder_prefix()

    class TenantUrlConf(ModuleType):
        urlpatterns = [
            path(
                f"{subfolder_prefix}/{tenant.domain_subfolder}/",
                include(urlconf),
            )
        ]

    return TenantUrlConf(tenant.domain_subfolder)
    
    get_subfolder_urlconf_override = get_subfolder_urlconf
  ```

### DIFF
--- a/django_tenants/middleware/subfolder.py
+++ b/django_tenants/middleware/subfolder.py
@@ -20,6 +20,7 @@ class TenantSubfolderMiddleware(TenantMainMiddleware):
     """
 
     TENANT_NOT_FOUND_EXCEPTION = Http404
+    get_subfolder_urlconf_override = None
 
     def __init__(self, get_response):
         super().__init__(get_response)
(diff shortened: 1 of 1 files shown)


isolationism:
Not sure how this will get solved upstream, but appreciate this workaround. 👍

sicter:
hi @atodorov, now sure what the best practice is around requesting reviews but wanted to get thoughts on this

PR #1166, opened 2025-06-30
Add support for SET LOCAL search_path via settings

mohamed-naser-awd:
**TENANT_USE_SET_LOCAL** (optional)
Type: bool
Default: False

If set to True, uses SET LOCAL search_path instead of SET search_path when setting the PostgreSQL schema for tenants.

This is primarily needed when using database connection poolers like PgBouncer in transaction pooling mode, where SET search_path is ineffective across transactions.

SET LOCAL ensures the schema change only applies within the current transaction, making it compatible with such environments.

Example:
TENANT_USE_SET_LOCAL = True

If not defined, the system defaults to False, using the standard SET search_path.


tomturner:
Needs the help written

PR #1192, opened 2025-11-10
Support Python 3.14

mikicz:


mikicz:
@tomturner I managed to fix the multi-processing executor on Python 3.14, tho I am not 100% happy with the fix - it's more a workaround...

kingbuzzman (approved):
The name `is_larger_than_314` seems odd, but other than that. LGTM

django_tenants/migration_executors/multiproc.py
obdulia-losantos, line 1: You only need to modify the init method and also do not require to check current python version

```python
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)

        # In Python 3.14 and above, the default start method is changed to 'forkserver',
        # which breaks the django migrations, raising django.core.exceptions.AppRegistryNotReady.
        # 'fork' is the previous default, which continues to work on Python 3.14.
        if multiprocessing.get_start_method() == 'forkserver':
            multiprocessing.set_start_method('fork', force=True)
```

mikicz:
#1233 is a nicer solution

PR #1213, opened 2026-02-26
Support Transaction Pool

mohamed-naser-awd:
Summary

This PR adds optional, first-class support for running django-tenants behind PgBouncer (or similar poolers) in transaction pooling mode, without leaking search_path between clients.

What’s included

- New setting: TENANT_PGBOUNCER_TRANSACTION_POOLING
  When True, django-tenants runs all work inside explicit transactions and sets the tenant schema with SET LOCAL search_path, so the schema is scoped to the current transaction only.
  The default is False, so existing behavior (session-level SET search_path) is unchanged.

- Backend changes (PostgreSQL backend wrapper)
  DatabaseWrapper now tracks whether search_path has been set in the current transaction and only runs SET LOCAL once per transaction.
  Autocommit-mode statements are transparently wrapped in BEGIN/COMMIT when pooling is enabled, so they still see the correct tenant schema.
  The tracking flag is cleared on _commit/_rollback to avoid leaking state across transactions.

- New utilities in django_tenants.utils
  get_pgbouncer_transaction_pooling() reads the new setting and drives the backend behavior.
  disable_transaction_pooling() context manager plus is_transaction_pooling_disabled() helper let callers opt out of the wrapping behavior for specific code paths (for example, raw SQL that manages its own transactions, or migrations that need different handling).

- New signal: tenant_transaction_began
  Emitted after SET LOCAL search_path runs at transaction start when pooling is enabled.
  Provides connection and schema_name so applications can hook in custom SQL that must run at the beginning of each transaction.

- Documentation updates (docs/use.rst)
  Adds a “PgBouncer transaction pooling” section explaining:
    - When to use this mode (only for transaction pooling, not session pooling).
    - How to enable TENANT_PGBOUNCER_TRANSACTION_POOLING and what changes in behavior.
    - The opt-out pattern using disable_transaction_pooling().
    - Important caveats, such as how run_on_commit behaves under synthetic transactions.

- Tests
  New PgBouncerTransactionPoolingTest coverage to verify:
    - Default (pooling off) behavior remains unchanged.
    - Autocommit queries succeed and run inside explicit transactions when pooling is on.
    - transaction.atomic() uses a single transaction and runs SET LOCAL once.
    - disable_transaction_pooling() correctly skips the wrapping behavior.
    - The internal _search_path_set_in_txn flag is cleared on commit.
    - tenant_transaction_began is emitted with the expected arguments.

Backward compatibility / risk

TENANT_PGBOUNCER_TRANSACTION_POOLING is off by default; existing deployments keep their current behavior.
When enabled, the behavior is opt-in, documented, and guarded by tests; callers can selectively opt out per code path via disable_transaction_pooling().

### DIFF
--- a/django_tenants/postgresql_backend/base.py
+++ b/django_tenants/postgresql_backend/base.py
@@ -6,7 +6,13 @@ from importlib import import_module
 from django.utils.module_loading import import_string
 
 from django_tenants.postgresql_backend.introspection import DatabaseSchemaIntrospection
-from django_tenants.utils import get_public_schema_name, get_limit_set_calls
+from django_tenants.signals import tenant_transaction_began
+from django_tenants.utils import (
+    get_limit_set_calls,
+    get_public_schema_name,
+    get_pgbouncer_transaction_pooling,
+    is_transaction_pooling_disabled,
+)
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ImproperlyConfigured, ValidationError
 import django.db.utils
@@ -69,6 +75,10 @@ class DatabaseWrapper(original_backend.DatabaseWrapper):
         self.search_path_set_schemas = None
         self.tenant = None
         self.schema_name = None
+        # PgBouncer transaction pooling: track if search_path was set in current transaction
+        # (Option B); cleared on _commit/_rollback.
+        self._search_path_set_in_txn = False
+        self._pgbouncer_wrapper_installed = False
         super().__init__(*args, **kwargs)
 
         # Use a patched version of the DatabaseIntrospection that only returns the table list for the
@@ -79,6 +89,9 @@ class DatabaseWrapper(original_backend.DatabaseWrapper):
 
     def close(self):
         self.search_path_set_schemas = None
+        self._search_path_set_in_txn = False
+        # Do not reset _pgbouncer_wrapper_installed so we don't append the wrapper
+        # again when the connection is reopened (execute_wrappers list is reused).
         super().close()
 
     def set_tenant(self, tenant, include_public=True):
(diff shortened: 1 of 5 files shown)


tomturner:
Thanks for putting this together — transaction-mode pooling is a real gap and `SET LOCAL` is the right instinct.

I'm closing this because, as written, the feature doesn't work: **all 6 of the tests added in this PR fail** on psycopg3. Checked out at 19c094e against Django 5.2 / psycopg 3.2.13 / PostgreSQL 17:

```
django.db.utils.ProgrammingError: can't change 'autocommit' now: connection in transaction status INTRANS

  File "django_tenants/postgresql_backend/base.py", line 284, in pgbouncer_execute_wrapper
    connection.set_autocommit(True)
  File "psycopg/_connection_base.py", line 260, in _check_intrans_gen
    raise e.ProgrammingError(...)

Ran 6 tests in 0.199s
FAILED (errors=6)
```

The cause is in `_make_pgbouncer_execute_wrapper`:

```python
connection.set_autocommit(False)
try:
    if connection.schema_name:
        with connection.cursor():   # runs SET LOCAL search_path -> opens a transaction
            pass
    result = execute(sql, params, many, context)
finally:
    connection.set_autocommit(True)  # <-- transaction still open here
```

`SET LOCAL search_path` plus the statement itself leave the connection at `INTRANS`, and psycopg refuses to flip `autocommit` while a transaction is in progress (psycopg2 raises here too). Nothing commits between the `execute` and the `finally`, so every autocommit-mode query raises — and the autocommit path is the main thing the feature is for.

If you'd like to pick it up again, the wrapper needs to commit (and roll back on the error path) before restoring autocommit, rather than relying on `set_autocommit(True)` to do it. Worth also reconsidering whether an execute wrapper is the right layer for this at all, versus overriding the transaction hooks — an execute wrapper runs per statement, so `BEGIN`/`COMMIT` per statement is a lot of round trips for something a pooler is meant to make cheaper.

Happy to look at a follow-up. Please make sure `./run_tests.sh` passes before re-opening.

PR #1215, opened 2026-03-02
Mention the use of context manager to activate a tenant schema

hongquan:


hongquan:
@tomturner please take a look.

PR #1216, opened 2026-03-24
Fix double `OutputWrapper`

dexbleeker:
This fixes the double use of `OutputWrapper`. 

When django-tenants wraps stdout in `OutputWrapper(sys.stdout)` with a style_func, and Django's `BaseCommand` wraps it again in `OutputWrapper(options["stdout"])`, the inner wrapper's default ending `"\n"` fires on the first write, splitting the output across two lines. By buffering until a real newline, the two writes are joined into one complete line before the prefix is applied.

Before:

```
[1/1 (100%) standard:tenant_name]   Applying app.0001_initial... 
[1/1 (100%) standard:tenant_name]  OK
```

After:

```
[1/1 (100%) standard:tenant_name]   Applying app.0001_initial... OK
```


srugano (approved):
Highly needed

PR #1263, opened 2026-08-26
Clear the cached search_path on savepoint rollback

### LINKED ISSUE #1262: TENANT_LIMIT_SET_CALLS: savepoint_rollback leaves a stale cached search_path
**Version:** 3.14.0 (reproduced on master `52b74e0`)
**Django:** 5.2.17 · **psycopg:** 3.3.4 · **PostgreSQL:** 16
**Only affects `TENANT_LIMIT_SET_CALLS = True`.**

### Summary

`_handle_search_path()` skips the `SET` while `search_path_set_schemas` is truthy. #1249 added a
`rollback()` override to clear that cache, with the right reasoning in the comment:

```python
@async_unsafe
def rollback(self):
    # A session-level `SET` is transactional in PostgreSQL: aborting the
    # transaction that issued it reverts the search_path. Forget the cached
    # value so the next cursor re-issues it rather than trusting a setting the
    # database has already discarded.
    self.search_path_set_schemas = None
    self._setting_search_path = False
    super().rollback()
```

`ROLLBACK TO SAVEPOINT` discards the setting the same way, and `savepoint_rollback()` is not
overridden. After a savepoint rollback the connection is on the schema the savepoint restored
while `search_path_set_schemas` still names the one that was set inside it — so the next cursor
sends no `SET` and the query runs against the wrong schema.

There is no error when the table exists in both schemas. It returns the other tenant's rows.

### Reproduction — the PostgreSQL behaviour

```
start:                        ('public',)
after SET inside savepoint:   ('pg_catalog',)
after ROLLBACK TO SAVEPOINT:  ('public',)     <-- reverted
```

### Reproduction — in this repo's test suite

Added next to `test_switching_search_path_limited_calls` in
`django_tenants/tests/test_tenants.py`:

```python
    @override_settings(TENANT_LIMIT_SET_CALLS=True)
    def test_savepoint_rollback_reissues_search_path(self):
        """`ROLLBACK TO SAVEPOINT` reverts a `SET` issued after the savepoint, so the
        cached search_path must not survive it -- otherwise the next cursor is skipped
        and the connection reads a schema it does not believe it is on."""
        tenant1 = get_tenant_model()(schema_name='tenant1')
        tenant1.save()

        domain1 = get_tenant_domain_model()(tenant=tenant1, domain='something.test.com')
        domain1.save()

        connection.set_schema_to_public()

        tenant2 = get_tenant_model()(schema_name='tenant2')
        tenant2.save()

        domain2 = get_tenant_domain_model()(tenant=tenant2, domain='example.com')
        domain2.save()

        connection.set_tenant(tenant1)
        DummyModel(name="tenant1 row").save()

        with transaction.atomic():
            with self.assertRaises(ValueError):
                with transaction.atomic():          # savepoint
                    connection.set_tenant(tenant2)  # SET is issued after the savepoint
                    DummyModel(name="tenant2 row").save()
                    raise ValueError

            # The savepoint rollback reverted search_path to tenant1's. Unless the cache
            # was cleared, no SET is issued here and this reads tenant1's row while the
            # connection reports tenant2.
            self.assertEqual(0, DummyModel.objects.count())

        self.created = [domain2, domain1, tenant2, tenant1]
```

On master:

```
FAIL: test_savepoint_rollback_reissues_search_path
AssertionError: 0 != 1
```

`connection.tenant` is `tenant2`; the row it counted belongs to `tenant1`.

`BaseTestCase` is a `TransactionTestCase`, so the outer `atomic()` is what makes the inner one a
savepoint.

### When it is reachable

Only when the schema switch itself happens inside the savepoint. `set_tenant()` clears the cache
without touching the database; the next cursor issues the `SET`. So the `SET` lands inside a
savepoint exactly when a `set_tenant()`/`set_schema()` call does.

The context managers are safe on their own, for two reasons worth spelling out:

- `schema_context.__exit__` calls `set_tenant()` on the way out, so an exception that escapes the
  block clears the cache before any enclosing `atomic()` rolls back.
- Entering `transaction.atomic()` opens a cursor, and that cursor issues the pending `SET` *before*
  the `SAVEPOINT` statement. So `with schema_context(x): with atomic(): ...` puts the `SET`
  outside the savepoint, and rolling it back changes nothing. (Verified: that shape passes on
  master with the flag on.)

What does reach it is a direct `connection.set_tenant()` / `set_schema()` inside an open
`atomic()` — which is what the test does, and what the library's own `run_migrations` does
(`set_schema(schema_name, include_public=False)`, then `set_schema(schema_name)`, then
`set_schema_to_public()`) whenever `TenantMixin.save()` provisions a schema inside a caller's
transaction. A failure between those calls leaves the connection reporting the new schema while
the rollback has put the session back on the old one.

### Suggested fix

```python
    @async_unsafe
    def savepoint_rollback(self, sid):
        super().savepoint_rollback(sid)
        # `ROLLBACK TO SAVEPOINT` discards a session-level `SET` issued after the
        # savepoint, exactly as `ROLLBACK` discards one issued after `BEGIN`. Forget the
        # cached value so the next cursor re-issues it.
        #
        # Cleared *after* the rollback, unlike in rollback(): obtaining the cursor for
        # `ROLLBACK TO SAVEPOINT` re-enters _cursor(), so clearing first would re-issue
        # the SET only for that same rollback to discard it -- leaving the cache stale.
        self.search_path_set_schemas = None
        self._setting_search_path = False
```

The ordering is the one difference from `rollback()`: psycopg's `rollback()` is a direct libpq
call, but `savepoint_rollback()` goes through `self.cursor()`, so clearing first would send a
`SET` for that same rollback to throw away.

With this applied, the full suite is 150 tests / 1 failure —
`files.test_storage.TenantFileSystemStorageTestCase.test_deprecated_module_raises_warning`, which
also fails on clean master and is unrelated.

### Context

Found while auditing a multi-tenant Django app that had just enabled `TENANT_LIMIT_SET_CALLS`
across ~1,300 tenant schemas. The one place it was live there was tenant provisioning under
`ATOMIC_REQUESTS`: a failure inside `run_migrations` left the connection reporting the new schema
while the session was back on `public`. The mismatch happened to be tenant↔public, so nothing
leaked — but the same mechanism between two tenant schemas returns another tenant's rows with no
error, which is why this seemed worth reporting rather than working around locally.

Happy to open a PR with the fix and the test if useful.

Amirshox:
Fixes #1262.

Follow-up to #1249. The `rollback()` override added there clears `search_path_set_schemas` because a `ROLLBACK` reverts the `SET`. `ROLLBACK TO SAVEPOINT` does the same thing to a `SET` issued after the savepoint, but `savepoint_rollback()` wasn't overridden. So with `TENANT_LIMIT_SET_CALLS` on, after a savepoint rollback the wrapper keeps skipping the SET and the connection is on a different schema than it reports.

I ran into this while turning the flag on in a ~1300-tenant app — details in the issue. Short version: `run_migrations` calls `set_schema()` directly, so if `TenantMixin.save()` runs inside a transaction and provisioning fails, the connection says it's on the new schema while the session is back on the old one. In my case the mismatch was tenant/public so nothing leaked, but between two tenant schemas it would return the wrong tenant's rows with no error.

One thing worth noting about the fix: the cache is cleared *after* `super().savepoint_rollback()`, not before like in `rollback()`. Getting the cursor for `ROLLBACK TO SAVEPOINT` goes through `_cursor()`, so clearing first just makes it issue a SET that the rollback immediately throws away.

The new test fails on master (`AssertionError: 0 != 1` — it counts tenant1's row while `connection.tenant` is tenant2) and passes with this. Rest of the suite is unchanged. `test_deprecated_module_raises_warning` fails for me on master as well, unrelated.


yarith28:
Hi @tomturner, please check this one out.

